PR #4, opened 2020-04-08
Update links to font tooling

rsheeter:


drott:
Sweet, thanks for going through the extra step!

PR #23, opened 2020-06-23
Rename colr-gradients-spec.md to README.md

davelab6:
So that it shows up inline on https://github.com/googlefonts/colr-gradients-spec


anthrotype:
I ended up renaming it back to colr-gradients-spec.md and creating a symlink to README.md, so that old links keep working.
https://github.com/googlefonts/colr-gradients-spec/commit/42e9d05b79321081fc7b1e881cc3b0ba91f8985b

PR #33, opened 2020-09-01
Change "radial" to "conical-radial"

### LINKED ISSUE #7: a radial gradient definition could be degenerate
I think normally the center of one circle (the gradient origin, stop position 0) would be **within** the circumference of the other circle (stop position 1). But the possibility exists that it is not: you could have two overlapping or even non-overlapping circles. 

So, what should be expected if that were the case? Do we need to treat that as invalid in OT, or just let different graphics libraries handle that however they will? (That could mean a font displays differently on different platforms.)

PeterConstable:
The proposed "radial" gradient is not a radial gradient in the strict sense but is a more general type of gradient that is elsewhere usually described as "conical" or "conical/radial". Calling it "radial" will lead to confusion.

Fixes #7

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -76,7 +76,7 @@ vector and fall back to no-color if the glyph is not found there either.
 A glyph using the new extension is mapped to an ordered list of layers.  Each
 layer includes a glyph id of shape as well as a *paint*.  There are three types
 of paint defined currently, with capacity for future extensions: solid, linear
-gradient, radial gradient.
+gradient, conical-radial gradient.
 
 We have added an "alpha" (transparency) scalar to each invocation of a palette
 color. This allows for the expression of translucent versions of palette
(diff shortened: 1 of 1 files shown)


behdad:
I've never seen "conical-radial". A conical gradient is different from what we have: that one the color is a function of the angle. That's not the case here.

https://css-tricks.com/snippets/css/css-conic-gradient/#:~:text=A%20conic%20gradient%20is%20similar,places%20them%20around%20the%20circle.

So, I don't know where you got the impression that you did.

PeterConstable:
What is proposed is sometimes referred to as "radial" but that only in a poor and confusing choice of terms. A radial gradient, strictly speaking, would have gradation of colour along rays emanating from a focal point.

CSS:
>A radial gradient is defined by a center point, an ending shape, and two or more color-stop points.

<https://developer.mozilla.org/en-US/docs/Web/CSS/radial-gradient#:~:text=A%20radial%20gradient%20is%20defined%20by%20a%20center,may%20be%20either%20a%20circle%20or%20an%20ellipse.>

Skia:
>Returns a shader that generates a radial gradient given the center and radius.

<https://skia-doc.commondatastorage.googleapis.com/doxygen/doxygen/html/classSkGradientShader.html#a39fc1395c51b358c4c34553baff82bdd>

What is proposed would really best be described as a "cylindrical" gradient, since the gradation of colours is along the length of a cylinder defined by circumferences of two circles. I've never seen that name used, however.

But I have seen this described as "conical".

Skia refers to the type of gradient proposed here as "two point conical":

>Returns a shader that generates a conical gradient given two circles

<https://skia-doc.commondatastorage.googleapis.com/doxygen/doxygen/html/classSkGradientShader.html#a602b579a185b4d1eff18b1572862b5f9>

The PDF spec refers to a "radial" shading but goes on immediate to describe it as conical.
>If neither circle contains the other, the shading depicts a cone. If the starting circle is larger, the cone appears to point out of the page. If the ending circle is larger, the cone appears to point into the page...

<https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdf_reference_archive/pdf_reference_1-7.pdf>

I'm not set on "conical-radial", but simply calling it "radial" is at best ambiguous, given clearly different definitions in use, and likely to be confusing. I'd be OK with Skia's term, "two-point conical".

davelab6:
@PeterConstable what do you want to do here? :)

PeterConstable:
I guess there's no undoing the existing ambiguous use in terminology. If we're going to refer to the Canvas definition for the gradient, it's probably least confusing to use the same term used there, but to describe it in terms that make clear what it really is like.

Issue #8, opened 2020-06-02
32-bit glyph IDs


PeterConstable:
It's unclear what the thinking was behind using 32-bit glyph IDs in the base glyph and layer records. (Note that v0 structures use uint16.)
* building in forward compatibility to a future in which fonts can have > 64K glyphs?
* keeping 4-byte-integral structures?
* other?

Note that this adds at least 6 bytes of padding to every colour glyph. For an emoji font, that will be an additional 12K to 20K (or more?) of padding.

(I'm not opposed, just want to understand the thinking behind proposing it, and to make sure the size implications are understood.)

behdad:
> * building in forward compatibility to a future in which fonts can have > 64K glyphs?

Yes, this was my intention.

> * keeping 4-byte-integral structures?

That's a non-issue to me since everyone has to support unaligned anyway.

> Note that this adds at least 6 bytes of padding to every colour glyph. For an emoji font, that will be an additional 12K to 20K (or more?) of padding.

Right.  But will definitely be a tiny bit of the font size, so I'm happy to "waste" that.  Compression takes care of it for any download situations as always.

rsheeter:
We care about the size at rest (e.g. in a system image) as well as the transfer size. I suggest we use `uint16`. If/when we are ready to introduce 32-bit gids we should make that as a global or spec wide change, perhaps introducing COLRv2 which is just v1 with a 4-byte offset.

anthrotype:
I agree with Rod, it feels odd to have 32-bit GlyphIDs only in COLRv1. We can always change it later and bump COLR table version.

rsheeter:
I propose to close this issue and pickup 32 bit gids and offsets as a standalone project, likely a W3C WG spawned from the shiny new CG. I imagine the end result of that is COLR gains the ability to support 16 or 32 bit gids, compiler can choose as appropriate for the specific font. @behdad would that work for you?

behdad:
Definitely works for me, yes. Just be more code and spec to maintain. Go ahead and close though.

davelab6:
Reopening and filing under the "COLRv2" milestone

alerque:
Relevant crosslink: https://github.com/MPEGGroup/OpenFontFormat/issues/10

rsheeter:
[...]

PR #34, opened 2020-09-03
First draft of PaintGroup (per Behdad proposal on #32)

### LINKED ISSUE #32: Multi-layer opacity
Compiling vector graphic sources, such as svgs, to COLR fonts will be significant to icon and emoji use cases (twemoji, noto emoji, and others come to mind). SVG allows application of opacity to multiple layers at once (by setting opacity on a group), meaning effectively render the group then apply opacity to the end result, as I understand it.

The current COLR spec doesn't allow for this. Building equivalent stacks of individually layers with opacity to match a group of layers with opacity appears tiresome. Also, when I tried to construct equivalent individual layers to a simple group opacity setup (courtesy of @anthrotype) by hand I immediately ran into halo effects (from anti-aliasing?). 

Off the top of my head there are three main options

1. Use some 2D graphics technique I'm unaware of to build equivalent single layers to match the group w/o introducing artifacts. Hooray, there is no actual problem :)
1. Add the ability to designate groups of layers with opacity to COLR
1. Accept that this is not supported by COLR; the set of things we can represent in COLR just got smaller

rsheeter:
Add reuse by group to COLRv1. Once acceptable will circle back to add blend mode.

Fixes #32.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -258,7 +258,7 @@ A group allows a set of layers to be composited independently and then have the
 final result composited into the end result as a single layer with alpha. This
 is similar to a `group` with `opacity` in an SVG.
 
-Concretely, a group reuses another COLR glyph with alpha and transform applied. For example:
+Concretely, a group reuses another glyph with alpha and transform applied. For example:
 
 1. A clock with marks at each hour
    - Define the first mark as a COLR glyph
@@ -267,6 +267,11 @@ Concretely, a group reuses another COLR glyph with alpha and transform applied. 
    - Define a glyph that has all twelve marks as above
    - Define a glyph for the hour-hand of the clock
    - Define N o’clock as the 12 marks untransformed, plus the hour-hand rotated
+
+If the `LayerV1Glyph` `gid` for the layer with paint format 4 (group) corresponds
+to a `BaseGlyphV1Record` then the layers of that glyph are taken as the layers in
+the group. If not, the referenced glyph is taken as a single layer with a solid paint
+in the text foreground color.
 
 
 # Structure of gradient COLR v1 extensions
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -20,6 +20,7 @@ December 2019
       - [Extend Reflect](#extend-reflect)
   * [Linear Gradients](#linear-gradients)
   * [Radial Gradients](#radial-gradients)
+- [Group](#group)
 - [Structure of gradient COLR v1 extensions](#structure-of-gradient-colr-v1-extensions)
 - [Implementation](#implementation)
   * [Font Tooling](#font-tooling)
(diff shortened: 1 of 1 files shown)


PeterConstable:
Thanks. It's not clear from the structures what is getting grouped by a format 4 paint. 

```
struct PaintGroup
{
  uint16              format; // = 4
  VarF2DOT14          alpha;
  Offset32<Affine2x3> transform; // May be NULL.
};
```
Did you intend to have something referring to other layers or glyphs? Does it group layers above it up to the next layer using format 4 paint? Or all layers above it?

Also, not sure if you saw my suggestion in #32 : I think it would be super useful to be able to define an opacity mask using a gradient --- similar to putting a BW gradient into an alpha channel or layer mask in Photoshop. Is that something that can be accommodated after circling back to add blend modes?

rsheeter:
> It's not clear from the structures what is getting grouped by a format 4 paint.

The idea is that for a layer with paint format 4 (group) the layer gid is to be interpreted as a reference to another COLR glyph. That glyph has a set of layers; those layers are the ones in the group. So, once you have a multi-layer COLR glyph you can reuse it - transformed - as part of another one.

> I think it would be super useful to be able to define an opacity mask using a gradient

I suggest we consider this along with blend modes.

PeterConstable:
Got it. I was forgetting the LayerV1Record that joins the paint to a gid. (Doh!)

rsheeter:
I added a bit more detail to the group section as I think it would be fair to say it was underspecified.

PeterConstable:
I've rev'd my OT formats draft accordingly: [rev 5](https://github.com/PeterConstable/OT_Drafts/blob/master/COLR_V1/COLRv1formats_rev5.md)

PeterConstable:
Any concern about burning GIDs?

(I know: would provide more impetus for 32-bit GIDs.)

behdad:
> Any concern about burning GIDs?

This will generally save GIDs because of the transformation possible. But valid point.

> (I know: would provide more impetus for 32-bit GIDs.)

Here's another idea:
1. reinstate 32bit GIDs in COLRv1 table,
2. allow GIDs within this table to go beyond the font's numGlyphs. Then all the group GIDs can use the top range without interference...

I know this is getting into dangerous territory. But this is what I envision a gradual shift to 32bit GIDs will look like. Kinda like how PC moved from 16bit pointers to 32bit by going through "High Memory", "Extended Memory", "Expanded Memory" etc address modes...

rsheeter:
Burning gids isn't any worse because of this. 

> a gradual shift to 32bit GIDs

Perhaps this is naive but ideally I would like to avoid this type of thing and instead attack 32 bit gids head-on as a separate project, hopefully drawing together interested parties from the nascent CG. I imagine that consists of defining a whole bunch of new table versions that support 32 bit gid/offsets.

rsheeter:
Merging per IM confirmation with Behdad this looks OK. If we want to debate 32 bit gids lets re-open #8.

PR #35, opened 2020-09-06
Support layer compositing and alpha

### LINKED ISSUE #13: add flag field to layer record?
In #3 LP mentions the "white on black" issue. He notes a mechanism that's already defined in the current COLR spec:

"A palette entry index value of 0xFFFF is a special case indicating that the text foreground color (defined by a higher-level client) should be used and shall not be treated as actual index into CPAL ColorRecord array."

And so suggests that an additional special case be defined in COLR, 0xFFFE, to mean "punch a hole in all previous layers".

Another approach that could be considered would be to add a separate flags field in the layer record. The flags could be used to indicate different blending options with previous layers, including "punch all previous layers", or "punch previous layer". (When punching, the separate paletteIndex could indicate a fill for the punched area?)

**Note:** I haven't thought through in detail what potential benefits are, or what problem would remain if not provided. (E.g., wouldn't it work to just add another non-punch layer on top?) The original issue that brought this up had to do with winding interactions of **overlapping contours within a single glyph**, not any interaction between different glyphs. So, I'm just throwing this out for discussion.

rsheeter:
Strawman for compositing support beyond just over. 

- Modes are as per https://www.w3.org/TR/compositing-1/
   - Most implementations support most modes, I have optimistically assumed that adding one or two isn't a deal breaker
      - D2D is notable in that it seems to have relatively few of the modes
      - sources
         - https://developer.apple.com/documentation/coregraphics/cgblendmode?language=objc
         - https://www.cairographics.org/operators/
         - https://docs.microsoft.com/en-us/windows/win32/api/d2d1_1/ne-d2d1_1-d2d1_composite_mode
         - https://source.chromium.org/chromium/chromium/src/+/master:third_party/skia/include/core/SkBlendMode.h
- Alpha support is added as a bonus composite mode
   - Allows use of another glyph as the alpha channel
   - This supports the desire for alpha gradients that has come up a few times

I attempted to follow the `PaintGroup` model and only force glyphs to "pay" bytes for compositing when actually used by making a new `Paint` instead of adding flags to the `LayerV1Record`.

Fixes #13.

Edit: the diff is a bit hard to read, https://github.com/googlefonts/colr-gradients-spec/blob/rs/colr-gradients-spec.md#compositing shows the updated md.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -466,7 +466,7 @@ struct PaintGroup
 struct PaintComposite
 {
   uint16              format; // = 5
-  BlendMode           blend;
+  Composite           composite;
   uint16              dest_gid;
 };
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -268,6 +268,20 @@ Concretely, a group reuses another glyph with alpha and transform applied. For e
    - Define a glyph for the hour-hand of the clock
    - Define N o’clock as the 12 marks untransformed, plus the hour-hand rotated
 
+The `LayerV1Glyph` `gid` field is resolved per [Self-referential gids](#self-referential-gids).
+
+# Composite
+
+By default layers are drawn on top of each other. Composite permits a variety of additional methods of combining layers.
+
+The compositing algorithms are specified by https://www.w3.org/TR/compositing-1/, with
+one exception: Alpha Channel. For alpha channel the dest colors are converted to the alpha channel for source by averaging the (R, G, B) parts.
+
+The `LayerV1Glyph` `gid` and the `PaintBlend` `dest_gid` are resolved per [Self-referential gids](#self-referential-gids).
+
+
+# Self-referential gids
+
 If the `LayerV1Glyph` `gid` for the layer with paint format 4 (group) corresponds
 to a `BaseGlyphV1Record` then the layers of that glyph are taken as the layers in
 the group. If not, the referenced glyph is taken as a single layer with a solid paint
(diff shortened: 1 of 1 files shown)


behdad:
Here's my counter-proposal:
```
struct PaintCompose
{
  uint16              format; // = 5
  Operator         op;
  Offset32<Paint> src;  
};
struct PaintMask
{
  uint16              format; // = 6
  Offset32<Paint> src;
  Offset32<Paint> mask;
};
```

rsheeter:
Spitballing, what about just making alpha optional on paint compose and having a single format?

```C++
struct PaintCompose
{
  uint16           format; // = 5
  Composite        mode;
  Offset32<Paint>  src;
  Offset32<Paint>  alpha; // May be NULL  
};
```

behdad:
Either should work.

Now let's double-check that these are enough to do the kind of flag styling (just the lighting part, not the waving) that I'm doing at noto-emoji:

- https://github.com/googlefonts/noto-emoji/blob/master/waveflag.c#L316
This draws a border around the flags, using operator=MULTIPLY and a solid color that is determined per flag at build time.

- https://github.com/googlefonts/noto-emoji/blob/master/waveflag.c#L344
Uses operator=SOFT_LIGHT to compose a flag-shaped gradient onto the flag to emulate lighting/shading.

So that seems alright to me.

Note that the layer gid is now acting as a clipping path for us. There's some unusual interactions about our setup that needs documentation. For example, the proposed PaintCompose / PaintMask can refer to simple Paint's, like Gradient easily, but if they refer to other PaintCompose, PaintMask, or PaintGroup, that needs to be clearly specified or disallowed.

All in all I like where this is going. I wonder if we should remove the layer-list idea and make it build a tree of operations instead. That's a drastic but simple change to make.  The only design constraint so far has been that:
1. color primitives are referenced from the CPAL table,
2. outlines are references by gid.

behdad:
cc @bungeman @drott @jfkthame

PeterConstable:
The notions "source" and "destination" may need some explaining, along with layer interaction. E.g., one way to understand the layers (probably the one most likely assumed by a reader) is that the first (bottom) layer is the destination, and the layers going up the stack are successively composed into it. So, if there's a struct with "dest_gid", then it won't be clear to the reader how the layers are supposed to interact.

behdad:
> The notions "source" and "destination" may need some explaining, along with layer interaction. E.g., one way to understand the layers (probably the one most likely assumed by a reader) is that the first (bottom) layer is the destination, and the layers going up the stack are successively composed into it. So, if there's a struct with "dest_gid", then it won't be clear to the reader how the layers are supposed to interact.

Yeah the `dest_gid` wasn't clear to me either.

`dest`, for blending purposes, is always the image built so far by composing of previous layers.

drott:
> * https://github.com/googlefonts/noto-emoji/blob/master/waveflag.c#L316
>   This draws a border around the flags, using operator=MULTIPLY and a solid color that is determined per flag at build time.
> * https://github.com/googlefonts/noto-emoji/blob/master/waveflag.c#L344
>   Uses operator=SOFT_LIGHT to compose a flag-shaped gradient onto the flag to emulate lighting/shading.

Mind if I ask one question before we discuss integrating such blending operators: The problem that Rod and Cosimo were facing originally regarding group opacity that solving this in preprocessing would have meant breaking cutting the overlapping parts of the groups into separate pieces, then performing complicated pre-computations for the gradient colors ahead of time, plus the fringing problems at the boundaries.

For the flag decoration cases you're describing for Noto Color Emoji - would we have similar difficulties in pre-processing? Or would it still be possible to cut (in a simpler way?) and map to the primitives we have at the moment?

Also: To understand the process better: So you start with a "flat" flag SVG and then perform filtering in cairo for waving them, giving them shadows and borders?

behdad:
For masking, only the alpha channel of the mask Paint will be used. This is unlike what Rod had written which was averaging the RGB (which is wrong at best).

PeterConstable:
Not sure how many options (R's original, B's alternative, R's counter) are still under consideration.

The two design options in the discussion both have PaintCompose wrapping a (simple?) paint and a compose mode. Every layer has _some_ default compose mode. So, this design involves adding a separate layer anytime a different compose mode is desired. What about putting the compose mode as a field in LayerV1Record?

behdad:
> The two design options in the discussion both have PaintCompose wrapping a (simple?) paint and a compose mode. Every layer has _some_ default compose mode. So, this design involves adding a separate layer anytime a different compose mode is desired. What about putting the compose mode as a field in LayerV1Record?

That is possible but would waste bytes for vast majority of layers that use operator=OVER.

behdad:
> Mind if I ask one question before we discuss integrating such blending operators: The problem that Rod and Cosimo were facing originally regarding group opacity that solving this in preprocessing would have meant breaking cutting the overlapping parts of the groups into separate pieces, then performing complicated pre-computations for the gradient colors ahead of time, plus the fringing problems at the boundaries.
> 
> For the flag decoration cases you're describing for Noto Color Emoji - would we have similar difficulties in pre-processing? Or would it still be possible to cut (in a simpler way?) and map to the primitives we have at the moment?
> 
> Also: To understand the process better: So you start with a "flat" flag SVG and then perform filtering in cairo for waving them, giving them shadows and borders?

Thanks for asking these questions Dominik.

In Noto Emoji, I started with flag SVG flags that I sourced from Wikipedia (https://github.com/behdad/region-flags); then yes, in the PNG version of the font, the referenced waveflag.c waves them and decorates them with a border and lighting.

The waving can be done as preprocessing (we haven't done yet). Note that the waving can in theory be done with SVG 2 as well but there's no working implementation of that. The border and lighting are much harder to do as preprocessing.  The result of, eg, operator=SOFT_LIGHT of two gradients CANNOT be specified as just another gradient...

drott:
> The waving can be done as preprocessing (we haven't done yet). Note that the waving can in theory be done with SVG 2 as well but there's no working implementation of that. The border and lighting are much harder to do as preprocessing. The result of, eg, operator=SOFT_LIGHT of two gradients CANNOT be specified as just another gradient...

Got it, but how about SOFT_LIGHT on top of a solid color? Can that be represented as new gradient (potentially as a result of preprocessing?)? I've been going through the flags and started wondering if there's any source flags that use a gradient? Also - it looks like the flag shading is done as a mesh source pattern - are we limited in this regard, or could this be mapped to a linear gradient in our case, clipped by the flag shape?

behdad:
> Got it, but how about SOFT_LIGHT on top of a solid color? Can that be represented as new gradient (potentially as a result of preprocessing?)? I've been going through the flags and started wondering if there's any source flags that use a gradient?

All Porter-Duff operators are linear, so yes for those. Of the separable blending modes, all except for color-dodge and color-burn look piecewise-linear if you fix the backdrop to a solid color, so those are also doable. The color-dodge / color-burn have a division by the source color so they can maybe approximated but not fully represented. The non-separable ones are definite not representable.

Same argument can be applied to all emoji. The SOFT_LIGHT etc can be used for shading emoji as well. And some emoji has been using multiple bands to emulate gradients.  Given that implementations need to be changed to implement COLRv1, and since the blend modes are readily available in pretty much all graphics systems, I think we should definitely use this opportunity to add them in.

> Also - it looks like the flag shading is done as a mesh source pattern - are we limited in this regard, or could this be mapped to a linear gradient in our case, clipped by the flag shape?

Correct. It can be approximated. The clipping is covered by the layer gid.

rsheeter:
> Given that implementations need to be changed to implement COLRv1, and since the blend modes are readily available in pretty much all graphics systems, I think we should definitely use this opportunity to add them in.

Vigorously agree.

To clarify, my original strawman was meant to allow expression of "combine A with B per rule C" to form a layer. Since A and B can in turn be defined as composites you can form a graph. Perhaps, as Behdad suggested above, if we want that we should replace layer list with a DAG with operations. Thats a relatively large change so IMO we should only do that if we have a clear need.

If we don't have a compelling reason to move away from the layer list then I think we are down to two options:

1. A single paint that has an explicit blend and a Paint for alpha
   - As Behdad noted, must clarify order of operations
1. Two separate paints, one to set blend and one to set alpha

We also need to address this:

> the proposed PaintCompose / PaintMask can refer to simple Paint's, like Gradient easily, but if they refer to other PaintCompose, PaintMask, or PaintGroup, that needs to be clearly specified or disallowed.

I would like to allow for more than just gradient, in particular to permit use of a painted shape to populate alpha. PaintGroup seems like a clear path to do so.

behdad:
> I would like to allow for more than just gradient, in particular to permit use of a painted shape to populate alpha. PaintGroup seems like a clear path to do so.

Can you tell me about your use-case for alpha, ala PaintGroup? I'm not sure that's the most useful thing.

What I see useful is to be able to apply a fade, blur, etc to a layer. PaintGroup becomes relevant for that because that's currently the only way we have to address more than one gradient. But I don't see it otherwise much useful because it because clipping more than masking for most glyphs.

rsheeter:
After VC about this we're now expecting a proposal for a revised structure using an explicit DAG; closing this PR in anticipation of that.

rsheeter:
New version in #37

PR #36, opened 2020-09-07
Refine group definition

drott:
Fix reference of LayerV1Glyph to LayerV1Record, add alpha computation
description.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -268,10 +268,15 @@ Concretely, a group reuses another glyph with alpha and transform applied. For e
    - Define a glyph for the hour-hand of the clock
    - Define N o’clock as the 12 marks untransformed, plus the hour-hand rotated
 
-If the `LayerV1Glyph` `gid` for the layer with paint format 4 (group) corresponds
-to a `BaseGlyphV1Record` then the layers of that glyph are taken as the layers in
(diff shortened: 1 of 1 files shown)


drott:
@rsheeter @behdad @PeterConstable PTAL.

PeterConstable (approved):

behdad:
I'm not sure about this:
```
The `alpha` value of the group is multiplied with the `alpha` values of the
individual `ColorIndex` values that are part of the `ColorLine` of the nested
`Paint` structures.
```

The whole point of a group paint is that the group is painted in isolation completely, and the result is alpha-blended with a fixed `alpha`. So the `alpha` and the individual colors never interact directly and no such multiplication happens ever.

drott:
I think I see what you mean. It is painted in isolation, based on ColorLine/ColorStops  including their individual alpha, then composited down to what's been painted so far, based on group alpha. But I agree it's indirect and left to the compositing implementation of the 2D graphics library.  Remove this part, or rephrase it, perhaps as follows:

> The  `alpha` value of the group is used for compositing down after isolated painting of the layers from `BaseGlyphV1Record` has completed.

rsheeter:
The offending section is deleted when we move to DAG (#37) per conversation yesterday so closing for now.

PR #37, opened 2020-09-10
Support layer compositing and alpha, mk2

### LINKED ISSUE #3: Consider CPAL/COLR for solving the white-on-black-with-overlaps issue
I would be pleased if another issue can be considered in this round of CPAL/COLR updates.

The “white on black” issue concerns colour-inverted b/w glyphs as seen in this image, from @adobe-fonts’s [white-on-black-vf](https://github.com/adobe-fonts/white-on-black-vf) repo, which contains a test font with this glyph.

67e45d00-818a-11e9-8166-bcc146b20892

_Top: actual behaviour. Bottom: desired behaviour_

As can be seen, the “white” or transparent parts of the glyph cannot use the overlapping contours that are useful in the “black“ or opaque parts of variable fonts.

In May 2019 @readroberts started a discussion on the otfontvar@unicode.org list (subject line _Need new path fill rule for variable fonts with white on black glyphs_), in which I suggested a COLR/CPAL approach. Later that month, @kenlunde posted the Adobe test font mentioned above to GitHub. In August 2019, Read summarized three possible approaches in an issue posted to the opentype-variations repo: [“White on Black” VF Rasterization Issue](https://github.com/OpenType/opentype-variations/issues/8):

1. a new rendering rule treating negative and positive winding differently
2. a COLR/CPAL solution
3. support for more complex graphics operations proposed by Renzhi Li

Read’s summary is well worth reading in full. He comes out firmly in favour of (1), a new rendering rule.

To me the rendering rule solution seems lined up for years of incompatibilities while rasterizers catch up, elegant only if we had a blank sheet. As Read acknowledges, it requires methods to export normal text glyphs as composited graphic objects for the vast number of system components that use OpenType fonts.

### Back to COLR/CPAL

It seems to me that COLR/CPAL is a pretty good solution. It is explicitly about layers. And it has a built-in fallback mechanism (to a b/w simple or composite glyph) for rasterizers that don’t support it (or that version of the tables). It lacks the ability to define masks, but we can spec that easily. An updated COLR table would define certain layers as "mask". We already have one special value in COLR, 0xFFFF:

> A palette entry index value of 0xFFFF is a special case indicating that the text foreground color (defined by a higher-level client) should be used and shall not be treated as actual index into CPAL ColorRecord array.  — _[COLR table specification](https://docs.microsoft.com/en-us/typography/opentype/spec/colr)_

**We can define 0xFFFE as the “mask” paletteIndex. Any layer “coloured” with paletteIndex 0xFFFE would punch holes in whatever had been built up so far.**

Typical use would be to have a single mask layer at the top of the stack, but it may be useful to have multiple mask layers at various positions in the stack.

COLR glyphs already piggy-back on top of the existing rasterizers, and this extra step seems to me less onerous than updating the core rasterizer. The SVG [`<clipPath>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/clipPath) element makes this technique easily translatable to SVG.

Read had three objections to the COLR solution, which I list here for consideration:

> The problem is really one of rasterization for normal text fonts. Since the ‘COLR’ table conveys the meaning of being a color font, it doesn’t seem to be the right place to be solving this problem, which is fundamentally a non-color one. Also, this could lead to problems with heuristics, such as font classification.

> There may also be some loss of effectiveness of anti-aliasing and hinting. Text renderers have been heavily optimized over the years for different environments. We suspect that some of this optimization could be lost when the ‘white’ paths are applied as a mask.

> Use of the ‘COLR’ table currently requires explicit opt-in on the part of the user, which is passed down through the API. This new use of the ‘COLR’ table would need to differentiate between the current use case, which is opt-in only, and the new use case, which involves the ‘COLR’ table being applied to the glyphs in question.

In respect of the anti-aliasing and hinting issue, it may be a worthwhile optimization for renderers, when processing COLR glyphs that _only_ use paletteIndices 0xFFFF (opaque) and 0xFFFE (transparent), to use a rasterizer path tuned for font rendering.

### LINKED ISSUE #13: add flag field to layer record?

rsheeter:
Take two (take 1 was #35) at layer compositing and alpha. Changes layer structure to a DAG based on chat with @behdad and @drott. 

- Modes are as per https://www.w3.org/TR/compositing-1/
   - Most implementations already support most modes
      - D2D is notable in that it seems to have relatively few of the modes
      - sources
         - https://developer.apple.com/documentation/coregraphics/cgblendmode?language=objc
         - https://www.cairographics.org/operators/
         - https://docs.microsoft.com/en-us/windows/win32/api/d2d1_1/ne-d2d1_1-d2d1_composite_mode
         - https://source.chromium.org/chromium/chromium/src/+/master:third_party/skia/include/core/SkBlendMode.h

Fixes #13.

The diff is a bit hard to read, https://github.com/googlefonts/colr-gradients-spec/blob/rs/colr-gradients-spec.md#compositing shows the updated md.

Edit: I believe this fixes #3 given that you get layers and `PaintComposite` can be used for hole-punching via [Source Out](https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcout).

Edit2: Note that I changed Extend and the Paint format to `uint8` instead of `uint16`; @anthrotype pointed out that I neglected to highlight this (ty).

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -85,7 +85,7 @@ are defined:
 1. **Transformed** reuses another paint, applying an affine transformation
 1. **Composite** reuses two other paints, applying a compositing rule to combine them
    * We draw on https://www.w3.org/TR/compositing-1/ for our composite modes
-1. **Glyph** draws a non-COLR glyph, filled using a type 1, 2, or 3 paint.
+1. **Glyph** draws a non-COLR glyph
    * A COLR v1 glyph made up of a list of Glyph paints is equivalent to a COLR v0 Layer Record with the added ability to use gradients.
 1. **Colr Glyph** reuses a COLR v1 glyph at a new position in the graph
 
@@ -468,12 +468,11 @@ struct PaintComposite
 // Paint a non-COLR glyph, filled as indicated by paint.
 // Akin to the COLRv0 Layer Record.
 // Clip paint to the region inked by gid
-// Paint must be unbounded
 struct PaintGlyph
 {
   uint8               format; // = 6
   uint16              gid;    // shall not be a COLR gid
-  Offset16<Paint>     paint;  // shall be format 1, 2, or 3
+  Offset16<Paint>     paint;
 }
 
 struct PaintColrGlyph
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -466,7 +466,7 @@ struct PaintGroup
 struct PaintComposite
 {
   uint16              format; // = 5
-  BlendMode           blend;
+  Composite           composite;
   uint16              dest_gid;
 };
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -22,6 +22,7 @@ December 2019
   * [Radial Gradients](#radial-gradients)
 - [Reusable Parts](#reusable-parts)
 - [Acyclic Graphs Only](#acyclic-graphs-only)
+- [Alpha](#alpha)
 - [Structure of gradient COLR v1 extensions](#structure-of-gradient-colr-v1-extensions)
 - [Implementation](#implementation)
   * [Font Tooling](#font-tooling)
@@ -274,6 +275,14 @@ The hour hand is reusable as a transformed glyph.
 Another example might be emoji faces: many have the same backdrop
 with different eyes, noses, tears, etc drawn on top.
 
+# Alpha
+
+The alpha channel for a layer can be populated using `PaintComposite`:
+
+- `PaintSolid` can be used to set a blanket alpha
+- `PaindLinearGradient` and `PaintRadialGradient` can be set gradient alpha
+- Mode [Source In](https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcin) can be used to mask
+
 # Acyclic Graphs Only
 
 `PaintColrGlyph` allows recursive composition of COLR glyphs. This
@@ -440,10 +449,6 @@ struct PaintTransformed
   Affine2x3           transform;
 };
 
-// Set alpha using Source In
-//    - a solid color can set uniform alpha
-//    - a gradient can set less uniform alpha
-//    - a glyph painted in alpha can be used to mask
 struct PaintComposite
 {
   uint8               format; // = 5
@@ -466,7 +471,6 @@ struct PaintColrGlyph
   uint8               format; // = 7
   uint16              gid;    // shall be a COLR gid
 }
-
 
 // Glyph root
 // NOTE: uint8 size saves bytes in most cases and does not
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -5,6 +5,7 @@ December 2019
 ### Authors:
 * Behdad Esfahbod ([@behdad](https://github.com/behdad))
 * Dominik Röttsches ([@drott](https://github.com/drott))
+* Roderick Sheeter ([@rsheeter](https://github.com/rsheeter))
 
 ## Table of Contents
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -81,7 +81,8 @@ COLR table is extended to expose a new vector of layers per glyph.  If a glyph
 is not found in the new vector, client will try finding it in the COLR v0 glyph
 vector and fall back to no-color if the glyph is not found there either.
 
-A glyph using the new extension is mapped to a list of layers. Each layers is formed by a directed acyclic graph of paints. Several different types of paint
+A glyph using the new extension is mapped to a list of layers. Each layer is
+formed by a directed acyclic graph of paints. Several different types of paint
 are defined:
 
 1. **Solid** paints a solid color
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -259,7 +259,7 @@ transforms.
 The alpha channel for a layer can be populated using `PaintComposite`:
 
 - `PaintSolid` can be used to set a blanket alpha
-- `PaindLinearGradient` and `PaintRadialGradient` can be set gradient alpha
+- `PaindLinearGradient` and `PaintRadialGradient` can be used to set gradient alpha
 - Mode [Source In](https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcin) can be used to mask
 
 # Reusable Parts
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -517,8 +517,6 @@ struct PaintComposite
 };
 
 // Paint a non-COLR glyph, filled as indicated by paint.
-// Akin to the COLRv0 Layer Record.
-// Clip paint to the region inked by gid
 struct PaintGlyph
 {
   uint8               format; // = 6
@@ -534,7 +532,7 @@ struct PaintColrGlyph
 
 // Glyph root
 // NOTE: uint8 size saves bytes in most cases and does not
-// preclude use of large layer counts via PaintComposite OVER
+// preclude use of large layer counts via PaintComposite.
 typedef ArrayOf<Offset16<Paint>, uint8> LayerV1List;
 
 // Each layer is OVER previous
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -268,6 +268,20 @@ Concretely, a group reuses another glyph with alpha and transform applied. For e
    - Define a glyph for the hour-hand of the clock
    - Define N o’clock as the 12 marks untransformed, plus the hour-hand rotated
 
+The `LayerV1Glyph` `gid` field is resolved per [Self-referential gids](#self-referential-gids).
+
+# Composite
+
+By default layers are drawn on top of each other. Composite permits a variety of additional methods of combining layers.
+
+The compositing algorithms are specified by https://www.w3.org/TR/compositing-1/, with
+one exception: Alpha Channel. For alpha channel the dest colors are converted to the alpha channel for source by averaging the (R, G, B) parts.
+
+The `LayerV1Glyph` `gid` and the `PaintBlend` `dest_gid` are resolved per [Self-referential gids](#self-referential-gids).
+
+
+# Self-referential gids
+
 If the `LayerV1Glyph` `gid` for the layer with paint format 4 (group) corresponds
 to a `BaseGlyphV1Record` then the layers of that glyph are taken as the layers in
 the group. If not, the referenced glyph is taken as a single layer with a solid paint
@@ -344,6 +358,52 @@ enum Extend : uint16
   EXTEND_REFLECT = 2,
 };
 
+// Compositing modes are taken from https://www.w3.org/TR/compositing-1/
+// NOTE: a brief audit of major implementations suggests most support most
+// or all of the specified modes.
+enum Composite : uint16
+{
+  // Porter-Duff modes
+  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators
+  COMPOSITE_CLEAR          =  0,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_clear
+  COMPOSITE_SRC            =  1,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_src
+  COMPOSITE_DEST           =  2,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dst
+  COMPOSITE_SRC_OVER       =  3,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcover
+  COMPOSITE_DEST_OVER      =  4,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstover
+  COMPOSITE_SRC_IN         =  5,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcin
+  COMPOSITE_DEST_IN        =  6,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstin
+  COMPOSITE_SRC_OUT        =  7,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcout
+  COMPOSITE_DEST_OUT       =  8,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstout
+  COMPOSITE_SRC_ATOP       =  9,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcatop
+  COMPOSITE_DEST_ATOP      = 10,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstatop
+  COMPOSITE_XOR            = 11,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_xor
+  COMPOSITE_PLUS           = 12,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_plus
+  COMPOSITE_SCREEN         = 13,  // https://www.w3.org/TR/compositing-1/#blendingscreen
+
+  // Blend modes
+  // https://www.w3.org/TR/compositing-1/#blending
+  COMPOSITE_OVERLAY        = 14,  // https://www.w3.org/TR/compositing-1/#blendingoverlay
+  COMPOSITE_DARKEN         = 15,  // https://www.w3.org/TR/compositing-1/#blendingdarken
+  COMPOSITE_LIGHTEN        = 16,  // https://www.w3.org/TR/compositing-1/#blendinglighten
+  COMPOSITE_COLOR_DODGE    = 17,  // https://www.w3.org/TR/compositing-1/#blendingcolordodge
+  COMPOSITE_COLOR_BURN     = 18,  // https://www.w3.org/TR/compositing-1/#blendingcolorburn
+  COMPOSITE_HARD_LIGHT     = 19,  // https://www.w3.org/TR/compositing-1/#blendinghardlight
+  COMPOSITE_SOFT_LIGHT     = 20,  // https://www.w3.org/TR/compositing-1/#blendingsoftlight
+  COMPOSITE_DIFFERENCE     = 21,  // https://www.w3.org/TR/compositing-1/#blendingdifference
+  COMPOSITE_EXCLUSION      = 22,  // https://www.w3.org/TR/compositing-1/#blendingexclusion
+  COMPOSITE_MULTIPLY       = 23,  // https://www.w3.org/TR/compositing-1/#blendingmultiply
+
+  // Modes that, uniquely, do not operate on components
+  // https://www.w3.org/TR/compositing-1/#blendingnonseparable
+  COMPOSITE_HSL_HUE        = 24,  // https://www.w3.org/TR/compositing-1/#blendinghue
+  COMPOSITE_HSL_SATURATION = 25,  // https://www.w3.org/TR/compositing-1/#blendingsaturation
+  COMPOSITE_HSL_COLOR      = 26,  // https://www.w3.org/TR/compositing-1/#blendingcolor
+  COMPOSITE_HSL_LUMINOSITY = 27,  // https://www.w3.org/TR/compositing-1/#blendingluminosity
+
+  // Use dest as the alpha channel for source
+  COMPOSITE_ALPHA_CHANNEL  = 28,
+};
+
 struct ColorLine
 {
   Extend             extend;
@@ -397,6 +457,15 @@ struct PaintGroup
   uint16              format; // = 4
   VarF2DOT14          alpha;
   Offset32<Affine2x3> transform; // May be NULL.
+};
+
+// The LayerV1Record gid is the source
+// The dest_gid is the dest
+struct PaintComposite
+{
+  uint16              format; // = 5
+  BlendMode           blend;
+  uint16              dest_gid;
 };
 
 // Header
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -21,6 +21,7 @@ December 2019
   * [Linear Gradients](#linear-gradients)
   * [Radial Gradients](#radial-gradients)
 - [Group](#group)
+- [Compositing](#compositing)
 - [Structure of gradient COLR v1 extensions](#structure-of-gradient-colr-v1-extensions)
 - [Implementation](#implementation)
   * [Font Tooling](#font-tooling)
@@ -268,16 +269,17 @@ Concretely, a group reuses another glyph with alpha and transform applied. For e
    - Define a glyph for the hour-hand of the clock
    - Define N o’clock as the 12 marks untransformed, plus the hour-hand rotated
 
-The `LayerV1Glyph` `gid` field is resolved per [Self-referential gids](#self-referential-gids).
+The `LayerV1Glyph` `gid` field is resolved per [self-referential gids](#self-referential-gids).
 
-# Composite
+# Compositing
 
-By default layers are drawn on top of each other. Composite permits a variety of additional methods of combining layers.
+By default layers are drawn on top of each other. Composite permits a variety of additional methods by specifying two glyphs to be combined according to a well defined
+algorithm.
 
 The compositing algorithms are specified by https://www.w3.org/TR/compositing-1/, with
 one exception: Alpha Channel. For alpha channel the dest colors are converted to the alpha channel for source by averaging the (R, G, B) parts.
 
-The `LayerV1Glyph` `gid` and the `PaintBlend` `dest_gid` are resolved per [Self-referential gids](#self-referential-gids).
+The `LayerV1Glyph` `gid` and the `PaintBlend` `dest_gid` are resolved per [self-referential gids](#self-referential-gids).
 
 
 # Self-referential gids
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -342,10 +342,11 @@ The following paints *may* be bounded:
 
 ## Bounding Box
 
-To simplify implementation allocation of drawing surface the
-`BaseGlyphV1Record` `gid` entry in the `glyf` table must provide
-the bounding box for the final COLR glyph by providing two
-on-curve points, one at `(min-x, min-y)` and one at `(max-x, max-y)`.
+To simplify implementation allocation of a drawing surface the
+bounding box of the glyph corresponding to the `BaseGlyphV1Record`
+should be taken to describe the drawing area for the COLR v1 glyph.
+
+Note: A `glyf` entry with two points at the diagonal extrema would suffice.
 
 # Structure of gradient COLR v1 extensions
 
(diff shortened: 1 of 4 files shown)


colr-gradients-spec.md
```diff
@@ -337,39 +353,73 @@ struct ColorStop
...
+  COMPOSITE_SRC_OVER       =  3,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcover
+  COMPOSITE_DEST_OVER      =  4,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstover
+  COMPOSITE_SRC_IN         =  5,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcin
+  COMPOSITE_DEST_IN        =  6,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstin
+  COMPOSITE_SRC_OUT        =  7,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcout
+  COMPOSITE_DEST_OUT       =  8,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstout
+  COMPOSITE_SRC_ATOP       =  9,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcatop
+  COMPOSITE_DEST_ATOP      = 10,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_dstatop
+  COMPOSITE_XOR            = 11,  // https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_xor
+  COMPOSITE_SCREEN         = 12,  // https://www.w3.org/TR/compositing-1/#blendingscreen
```
anthrotype, line 382: the https://www.w3.org/TR/compositing-1/ spec lists "screen" under the blend modes rather than a porter-duff compositing operators. The way you split the two sections implies it belongs to the latter.

colr-gradients-spec.md
```diff
@@ -337,39 +353,73 @@ struct ColorStop
   ColorIndex color;
 };
 
-enum Extend : uint16
+enum Extend : uint8
```
anthrotype, line 356: i'm ok to save 1 byte here and in the paint formats (why where they uint16 to begin with?).
Maybe you could make changes like these to pre-existing types more prominent in the main PR message, as that will help implementers (me) updating existing draft implementations.

colr-gradients-spec.md
```diff
@@ -392,23 +442,42 @@ struct PaintRadialGradient
...
+  uint8               format; // = 5
+  CompositeMode       mode;
+  Offset16<Paint>     src;
+  Offset16<Paint>     backdrop;
 };
 
-typedef ArrayOf<LayerV1Record, uint32> LayerV1List;
+// Clip paint to the region inked by gid
+// Paint must be unbounded
+struct PaintGlyph
```
anthrotype, line 462: what is the use of `PaintGlyph`? I don't see it presented or discussed anywhere.

colr-gradients-spec.md
```diff
@@ -252,27 +254,41 @@ inverse-transform approach noted above can fall back to a linear-gradient
 combined with a clipping path to achieve proper rendering of problematic affine
 transforms.
 
-# Group
+# Alpha
 
-A group allows a set of layers to be composited independently and then have the
-final result composited into the end result as a single layer with alpha. This
-is similar to a `group` with `opacity` in an SVG.
+The alpha channel for a layer can be populated using `PaintComposite`:
```
anthrotype, line 259: I don't see where this "alpha channel for a layer" is defined.. I suspect I'm missing something basic about these compositing and blending modes.

anthrotype (commented):
I left a few comments below.

Incidentally, the "High-level Design" section at the top needs updating as it currently says "There are four types of paint defined currently..." whereas this PR defines more.

colr-gradients-spec.md
```diff
@@ -252,27 +254,41 @@ inverse-transform approach noted above can fall back to a linear-gradient
 combined with a clipping path to achieve proper rendering of problematic affine
 transforms.
 
-# Group
+# Alpha
 
-A group allows a set of layers to be composited independently and then have the
```
anthrotype, line 257: without the PaintGroup how do we solve the problem of multi-layer opacity discussed #32?

colr-gradients-spec.md
rsheeter: Will do, good catch.

colr-gradients-spec.md
rsheeter: It basically replaces LayerV1Record: paint a non-color glyph using the specified paint. I will update the comment to hopefully be more informative.

colr-gradients-spec.md
rsheeter: Via the A part of the colors used. [Source in](https://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcin) defines color exclusively in terms of the source color while combining alpha so you can do things like combine my source with a solid paint whose alpha is <value> or do fancy things like combine with a gradient whose stops have different alphas.

colr-gradients-spec.md
rsheeter: You would use `PaintComposite` to combine `PaintColrGlyph` with a `Paint` that populates alpha.

colr-gradients-spec.md
rsheeter: Fixed, ty.

rsheeter:
Updated high level design, ty @anthrotype for noticing I hadn't yet.

colr-gradients-spec.md
anthrotype: ah, so you mean for PaintSolid you set it from the A of the RGBA color in the CPAL table. That means it won't be variable, unlike the alpha value for the gradient paints.

colr-gradients-spec.md
```diff
@@ -76,10 +76,18 @@ COLR table is extended to expose a new vector of layers per glyph.  If a glyph
 is not found in the new vector, client will try finding it in the COLR v0 glyph
 vector and fall back to no-color if the glyph is not found there either.
 
-A glyph using the new extension is mapped to an ordered list of layers.  Each
-layer includes a glyph id of shape as well as a *paint*.  There are four types
-of paint defined currently, with capacity for future extensions: solid, linear
-gradient, radial gradient, and group.
+A glyph using the new extension is mapped to a list of layers. Each layers is formed by a directed acyclic graph of paints. Several different types of paint
```
anthrotype, line 79:
```suggestion
A glyph using the new extension is mapped to a list of layers. Each layer is formed by a directed acyclic graph of paints. Several different types of paint
```

colr-gradients-spec.md
anthrotype: Ah, I didn't notice you also removed the LayerV1Record.
I am a bit confused by the characterization of PaintGlyph and PaintColrGlyph structures as "paints" instead of just "layers". In earlier drafts and so far, we were thinking of a layer as a structure connecting a glyph (via its gid) and some paint (a color, or gradient). Now the distinction between layer and paint seems to be blurred.
Early on you say that base glyphs are mapped to "list of layers", however the LayerV1List is now an array of offsets to Paint structures, not layers strictly speaking. 

One question. What happens if these offsets in LayerV1List is directly one of the old format 1, 2 and 3 "pure" or "unbounded" Paints that don't have a glyph index associated with them? Should we disallow these in the top level and require they be children of PaintGlyph?

rsheeter: I've been deliberately light on restrictions. I suspect we do need to add some, but hopefully biasing towards having as few as possible. 

I don't think you want to require the unbounded paints be children of PaintGlyph; PaintComposite {PaintGlyph, mode, Paint Solid or Gradient} should be valid. I'm not completely clear if the extent to fill can be given a clear definition when an unbounded paint is never bounded/clipped.

@behdad @drott wdyt?

behdad:
The part saying:
```
Glyph draws a non-COLR glyph, filled using a type 1, 2, or 3 paint.
```
I don't think the limitation is necessary.

behdad:
Looks good to me. I think the part about bounded vs unbounded paints needs more work.

To answer Cosimo's comment: in the new design we have one type of thing, called Paint. It's the same as layer, or image. Any paint is a full image. It is either bounded or unbounded. The following needs incorporation into the spec:

- Each layer in a `LayerV1List` must be bounded.

- Solid and gradient paints are unbounded.

- A PaintTransformed is bounded if and only if the source is bounded,

- PaintGlyph and PaintColrGlyph results are always bounded. The source for PaintGlyph does *not* have to be unbounded. So that needs fixing in the current PR.

- PaintComposite result's boundedness is a bit more complicated as it's a function of the boundedness of src & backdrop as well as the CompositeMode. This is for example addressed similarly in the cairo documentation, although only in relationship to the source (if I recall correctly): https://www.cairographics.org/operators/

I try to enumerate what's needed to be speced here (please double-check my work):

The following CompositeModes are always bounded: CLEAR.
The following CompositeModes are bounded if and only if src is bounded: SRC, SRC_OVER, SRC_OUT.
The following CompositeModes are bounded if and only if backdrop is bounded: DEST, DEST_OVER, DEST_OUT.
The following CompositeModes are bounded if and only if *either* src or backdrop are bounded: SRC_IN, DEST_IN.
All the other CompositeModes are bounded if and only if *both* src or backdrop are bounded.

rsheeter:
@behdad updated per feedback, PTAL. My list of boundedness differs only for SRC/DEST_OVER, which I have as bounded if src *and* backdrop are bounded.

While I'm at it, summon @anthrotype @drott!

drott:
I did a longer first pass today and here's some comments, first the important ones which I think we need to address in this CL first, then some considerations we may want to do in subsequent edits. At the bottom, I'll be pasting a sketch of the algorithm I have in mind for drawing this.

#### For this CL:
1. I think we need a pre-computed bounding box. After discussing with Cosimo we came to the following idea: Mac COLR v0 implementation uses two points at the extremes in the `glyf` table to size the allocation. Cosimo and I considered doing the the same here. Background: For COLR v0 and COLR v1 so far without transforms and blend modes, the union of layers suffices, and was rather fast to compute. For v1 with the edits in this CL, the root layer extents are not easy to compute any more. We need to describe what should be the top level surface extent, and how to retrieve it from the gid entry of the `glyf` table and use a rectangle that fits the extreme. Ideally that rectangle should be tight around the color glyph shape, as it controls allocation size.
2. There is a mismatch between `PaintComposite`'s terminology: **src** and **backdrop** and the blending modes terminology: **src** and **destination**. We should make that consistent to use only one or the other.

#### In subsequent edits:

1. Offsets of various `PaintXYZ`: Do these offsets still refer to the beginning of the parent struct, or do we want all of them to point to a common reference point for one COLR v1 glyph? If it's a graph, to me it feels slightly confusing to keep track of the parents, but might be subjective - what do others think?
2. Reuse of individual `PaintGlyph`, `PaintTransformed`, `PaintComposite` etc: Can one COLR v1 glyph refer to PaintGlyph "objects" that come from a different COLR v1 glyph? Or is our "maximum" unit of reuse `PaintColrGlyph`?
3. In Chrome we support glyph intercepts for ink skipping in the underline: Can these still be efficiently computed with path operations, or only from looking at which pixels are transparent on the root layer?
4. So far `PaintGlyph` aways has _ClipIn_ semantics, i.e. fill the inside of the shape with what's referenced as Paints from PaintGlyph: Do we need _ClipOut_ semantics? - Current thinking: No, probably not, as you could achieve the same by subtracting the desired shape from the extent box.
5. We should rephrase "High Level Design" to also talk about how we model that to available primitives in a 2D graphics library, then perhaps regroup Transforms and Composites under the heading "Graphical Primitives" and potentially rename the section. I can do that in subsequent edits.

#### Sketch of algorithm (perhaps include this in a prose section of the spec?), feedback or corrections welcome:

```
Allocate a bitmap for the glyph according to glyf table entry extends for gid
0) Traverse layers for gid, add current gid to blacklist *1
 a) Paint a paint, switch:
    1) PaintGlyph
         gid must not COLRv1
         saveLayer
         setClipPath to gid path
           recurse to a)
         restore
    2) PaintColorGlyph
         gid must be from 
         if gid on recursion blacklist, do nothing
         recurse to 0) with different gid
    3) PaintComposite
          paint Paint for backdrop, call a)
          saveLayer() with setting composite mode, on SkPaint
          paint Paint for src, call a)
          restore with save composite mode
    4) PaintTransformed
          saveLayer()
          apply transform 
          call a) for paint
          restore
    5) PaintRadialGradient
          SkCanvas::drawPaint with radial gradient configured
          (expected to be bounded by parent composite mode or clipped by current clip, check bounds?)
    6) PaintLinearGradient
          SkCanvas::drawPaint with liner gradient configured
          (expected to be bounded by parent composite mode or clipped by current clip, check bounds?)
    7) PaintSolid
          SkCanvas::drawColor with color configured

 *1 (in the implementaiton, potentially collapse each pair as PaintComposite (3) with first gid as backdrop, second as src, or just draw on root SkCanvas directly)    
```

behdad:
> My list of boundedness differs only for SRC/DEST_OVER, which I have as bounded if src _and_ backdrop are bounded.

Correct. Thanks.

rsheeter:
Ty @drott! I have some edits to make (once meeting hell ends); a few immediate comments:

> 1. I think we need a pre-computed bounding box

I like this idea, will update.

> There is a mismatch between PaintComposite's terminology: src and backdrop and the blending modes terminology: src and destination.

https://www.w3.org/TR/compositing-1 refers to backdrop. If this is OK to cite from the font spec then I suggest we use the term backdrop. Does that seem reasonable?

> 2. Reuse of individual `PaintGlyph`, `PaintTransformed`, `PaintComposite` etc: Can one COLR v1 glyph refer to PaintGlyph "objects" that come from a different COLR v1 glyph? Or is our "maximum" unit of reuse `PaintColrGlyph`?

Reuse any paint the font compiler wants. I think of the DAG as picking parts from a common pool.

behdad:
@rsheeter please add yourself to the authors as well. It feels weird every time opening your draft and seeing my name and Dominik's there only.

behdad:
Thanks Dominik. Replying to some of the easier ones:

> 1. Offsets of various `PaintXYZ`: Do these offsets still refer to the beginning of the parent struct, or do we want all of them to point to a common reference point for one COLR v1 glyph? If it's a graph, to me it feels slightly confusing to keep track of the parents, but might be subjective - what do others think?

The offsets are from the beginning of the current paint.

> 2. Reuse of individual `PaintGlyph`, `PaintTransformed`, `PaintComposite` etc: Can one COLR v1 glyph refer to PaintGlyph "objects" that come from a different COLR v1 glyph? Or is our "maximum" unit of reuse `PaintColrGlyph`?

Yes they can, by way of byte reuse. That's the beauty of this approach: the builder doesn't need to even identify duplicate paints. The compiler reuses bytes that are similar.

> 3. In Chrome we support glyph intercepts for ink skipping in the underline: Can these still be efficiently computed with path operations, or only from looking at which pixels are transparent on the root layer?

It can *if* you define what the shape of a gradient is. Or assume gradients to be opaque and then is not much different from what you currently do; just additional transform...  But yeah in this case pixel check may be easier.

> 4. So far `PaintGlyph` aways has _ClipIn_ semantics, i.e. fill the inside of the shape with what's referenced as Paints from PaintGlyph: Do we need _ClipOut_ semantics? - Current thinking: No, probably not, as you could achieve the same by subtracting the desired shape from the extent box.

Correct.

> 5. We should rephrase "High Level Design" to also talk about how we model that to available primitives in a 2D graphics library, then perhaps regroup Transforms and Composites under the heading "Graphical Primitives" and potentially rename the section. I can do that in subsequent edits.

Sgtm.

> 6. Do we want to keep the `LayerV1List` or simplify this to using `PaintComposite` at the root already? Feels slightly redundant but may be useful as a storage optimisation.

What Rod said. The list of layers is also more efficient since new layer can be painted right on top of the existing output. Same might be doable with the DAG but needs more analysis to implement that way.

colr-gradients-spec.md
rsheeter: Behdad responded below, and I believe we've come to agreement on boundedness; resolving. As always, please reopen if not adequately addressed.

rsheeter:
Added a note about base glyf bbox, a note about offsets being relative to the containing struct, and captured @drott rendering pseudocode.

I believe I have addressed all of @drott notes except the one that includes the explicit offer to make the changes :)

> 5. We should rephrase "High Level Design" to also talk about how we model that to available primitives in a 2D graphics library, then perhaps regroup Transforms and Composites under the heading "Graphical Primitives" and potentially rename the section. I can do that in subsequent edits.

behdad:
> 1. I think we need a pre-computed bounding box. After discussing with Cosimo we came to the following idea: Mac COLR v0 implementation uses two points at the extremes in the `glyf` table to size the allocation. Cosimo and I considered doing the the same here. Background: For COLR v0 and COLR v1 so far without transforms and blend modes, the union of layers suffices, and was rather fast to compute. For v1 with the edits in this CL, the root layer extents are not easy to compute any more. As graphics libraries may want to know what size to allocate in the glyph cache, we need to describe what should be the top level surface extent, and how to retrieve it from the respective gid entry of the `glyf` table and use a rectangle spanned by the extremes. Ideally that rectangle should be tight around the color glyph shape, as it controls allocation size.

That's a very good point.

Can we spec that the `glyf`/otherwise bounding box of the *first* layer is used? And then have a Note that a degenerate outline with two diagonal extrema points can be used in the first layer glyph to extend the bounding box as necessary?

behdad:
> Can we spec that the `glyf`/otherwise bounding box of the _first_ layer is used? And then have a Note that a degenerate outline with two diagonal extrema points can be used in the first layer glyph to extend the bounding box as necessary?

The alternative is to say that the `glyf`/otherwise bounding box of the *current* gid is used. That's what `sbix` does. Now I realize that that's probably what @drott meant. Let me read @rsheeter's change to see if that's clear or needs clarification.

colr-gradients-spec.md
```diff
@@ -335,8 +339,17 @@ The following paints *may* be bounded:
    - Bounded IFF src AND backdrop are bounded
       - *all other modes*
 
+## Bounding Box
+
+To simplify implementation allocation of drawing surface the
+`BaseGlyphV1Record` `gid` entry in the `glyf` table must provide
```
behdad, line 346: This phrase: "`BaseGlyphV1Record` `gid` entry in the `glyf`" doesn't make sense and needs fixing. Removing the `BaseGlyphV1Record` might do.

colr-gradients-spec.md
```diff
@@ -335,8 +339,17 @@ The following paints *may* be bounded:
...
+the bounding box for the final COLR glyph by providing two
```
behdad, line 347: The "by providing" should become a note, not a recipe.

behdad: > not a recipe

Or, "not a requirement" I meant.

colr-gradients-spec.md
```diff
@@ -74,10 +81,18 @@ COLR table is extended to expose a new vector of layers per glyph.  If a glyph
...
-layer includes a glyph id of shape as well as a *paint*.  There are four types
-of paint defined currently, with capacity for future extensions: solid, linear
-gradient, radial gradient, and group.
+A glyph using the new extension is mapped to a list of layers. Each layers is formed by a directed acyclic graph of paints. Several different types of paint
+are defined:
+
+1. **Solid** paints a solid color
+1. **Linear gradient** paints a linear gradient
+1. **Radial gradient** paints a radial gradient
+1. **Transformed** reuses another paint, applying an affine transformation
```
anthrotype, line 90: I don't remember if this has been already discussed, but what happens if PaintTransformed references one of the unbounded paints (formats 1, 2, 3)?
At least format 1 (solid paint) has no actual (x, y) coordinates to be transformed.
For linear and radial gradient paint, one could maybe apply the transformation to the gradient points.
Radial also has its own Affine2x2. Would one concatenate this with PaintTransformed?

Or was the idea that PaintTransformed be only used in combination with PaintGlyph or PaintColorGlyph?

colr-gradients-spec.md
drott: I'd phrase that a bit more strict, perhaps:

> The surface extents on which to paint the `Paint` operations of a `LayerV1List` is defined by two on-curve points in the `glyf` table at the `gid` entry for the same `gid` as in this glyph's `BaseGlyphV1Record`: one at `(min-x, min-y)` and one at `(max-x, max-y)`.

@behdad:

> Can we spec that the glyf/otherwise bounding box of the first layer is used? And then have a Note that a degenerate outline with two diagonal extrema points can be used in the first layer glyph to extend the bounding box as necessary?

Your suggestion to spec'ing is as "bounding box of first layer": Did you mean without traversing all paints  requiring that the first paint of the first layer is bounded?

colr-gradients-spec.md
```diff
@@ -392,23 +501,43 @@ struct PaintRadialGradient
...
+  uint8               format; // = 5
+  CompositeMode       mode;
+  Offset16<Paint>     src;
+  Offset16<Paint>     backdrop;
 };
 
-typedef ArrayOf<LayerV1Record, uint32> LayerV1List;
+// Paint a non-COLR glyph, filled as indicated by paint.
+// Akin to the COLRv0 Layer Record.
+// Clip paint to the region inked by gid
```
drott, line 521: Either keep
"// Paint a non-COLR glyph, ..."
and drop
"Clip paint to the region inked by gid"

Or perhaps replace with
"// Paint a non-COLR glyph by clipping to region inked by ink, filled as indicated by paint."

Nit/Minor: Not sure the "Akin to the COLRv0 Layer Record." comment helps.

colr-gradients-spec.md
drott: > The alternative is to say that the `glyf`/otherwise bounding box of the current gid is used.

Aren't we looking at `glyf` for the "current" gid? Yes, I did mean SBIX style, if I understood it correctly.

drott:
Some phrasing suggestions in previous comments, mostly looks good to me.

#### Re previous parts of the conversation:

> Terminology: destination / backdrop.

I see that in the W3C spec also both terms are used. I would like more consistency but okay for me to have `backdrop` as the struct member. Doesn't look to good to rename all operators to `..._BACKDROP_`.

> My question regarding whether to keep layer layerlist.

Makes sense, ok to keep as more compact representation than a chain of `PaintComposite` + OVER operators.

> > 1. Reuse of individual `PaintGlyph`, `PaintTransformed`, `PaintComposite` etc: Can one COLR v1 glyph refer to PaintGlyph "objects" that come from a different COLR v1 glyph? Or is our "maximum" unit of reuse `PaintColrGlyph`?
> 
> Yes they can, by way of byte reuse. That's the beauty of this approach: the builder doesn't need to even identify duplicate paints. The compiler reuses bytes that are similar.

Ok, since the offsets are from the current paint, that would mean that the compiler would have to move the reused ones to the end, so that they all can point forward, right?

colr-gradients-spec.md
drott: I think the algorithm covers that: a `PaintTransformed` creates a layer, then applies the transformation matrix. Then child paints happen. If the child paint is a `PaintSolid`, `PaintLinearGradient`, `PaintRadialGradiant` - these would be implemented by calling Skia's `drawPaint` which fills "everything" at this stage with the fill. The gradient matrices (for radial) are set up with the point info taken as is from the struct. (the points inverted with that matrix). So effectively, yes, the transformations of outer `PaintTransformed` and inner gradient matrices are combined. 

Is that the right understanding, @behdad?

colr-gradients-spec.md
rsheeter: Fixed.

colr-gradients-spec.md
```diff
@@ -252,30 +267,90 @@ inverse-transform approach noted above can fall back to a linear-gradient
...
+`PaintColrGlyph` allows recursive composition of COLR glyphs. This
+is desirable for reusable parts but introduces the possibility of
+a cyclic graph. Implementations should track the COLR gids they have
+seen in processing and fail if a gid is reached repeatedly.
+
+## Bounded Layers Only
+
+Every entry in the `LayerV1List` must define a bounded region.
+Implementations must confirm this invariant. Unbounded layers must
+not render.
```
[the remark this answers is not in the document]
rsheeter: I tried that but it feels as redundant given the section starts with "Every entry in the `LayerV1List` must define a bounded region."

Alternate phrasing? OK to leave unchanged?

colr-gradients-spec.md
rsheeter: I meant to take the bounding box for the glyph whose id is in the BaseGlyphV1Record::gid field. Is that the same as the "current" gid?

I dislike having to process the first layer; removes the benefit of a simple path to knowing the extents.

I will move the two on-curve points bit to be a note.

colr-gradients-spec.md
rsheeter: Updated, better?

colr-gradients-spec.md
rsheeter: Fixed

colr-gradients-spec.md
```diff
@@ -342,10 +343,11 @@ The following paints *may* be bounded:
 
 ## Bounding Box
 
-To simplify implementation allocation of drawing surface the
-`BaseGlyphV1Record` `gid` entry in the `glyf` table must provide
-the bounding box for the final COLR glyph by providing two
-on-curve points, one at `(min-x, min-y)` and one at `(max-x, max-y)`.
+To simplify implementation allocation of a drawing surface the
+bounding box of the glyph corresponding to the `BaseGlyphV1Record`
```
drott, line 347: I would perhaps insert "bounding box of the `glyf` table glyph" to make clear that we're looking at the other table and not mention that only in the note.

colr-gradients-spec.md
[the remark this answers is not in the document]
drott: Ok to leave unchanged.

drott:
LGTM to merge - I will follow-up with some section re-arrangements in a separate change.

colr-gradients-spec.md
anthrotype: further below, in "Bounded Layers Only" section, where it says

> Every entry in the `LayerV1List` must define a bounded region... Unbounded layers must
not render... `PaintTransformed` is bounded IFF the source is bounded

so a PaintTransformed with an unbounded paint the top level will not render -- unless if it's in turn bounded by a parent PaintGlyph or PaintComposite -- Is that correct?

drott: Yes, as discussed with Behdad: We would keep a flag on layer boundedness and collapse that upwards. If the top most layer is unbounded, don't draw.

PR #53, opened 2020-09-29
Remove radial gradient 2x2, clarify PaintTransformed interaction with gradients

### LINKED ISSUE #45: Transform in linear/radial gradient redundant?
Originally we added an `Affine2x2` to linear and radial gradients because necessary. But since we have added an `Affine2x3` now, I believe the `Affine2x2`s can be removed.

rsheeter:
Also remove the special rules for handling that 2x2.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -215,14 +215,14 @@ be rendered.
 
 ## Radial Gradients
 
+Radial gradients in this proposal are defined based on circles. If subject to
+a transform (via `PaintTransformed`) those circles may become ellipses.
+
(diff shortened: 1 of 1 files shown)


behdad:
Awesomeness. LGTM. Thanks!

rsheeter:
Per IM, working on expanding this to adjust linear gradient as well.

behdad:
We need the actual points P0 and P1 or variations cannot work correctly.

rsheeter:
I'm going to back out the linear changes and put those in a separate PR then.

PR #55, opened 2020-09-29
Remove P2 from linear gradient; just use PaintTransformed

### LINKED ISSUE #45: Transform in linear/radial gradient redundant?

rsheeter:
P2 is effectively a transform. Just use a real transform now we have one.

Combined with #53, Fixes #45

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -202,12 +202,7 @@ defining stripes in rotating colors.
 ## Linear Gradients
 
 We propose definitions of linear gradients with two color line points P0 and P1
-between which a gradient is interpolated. A point P2 is defined to rotate the
-gradient angle / orientation separately from the color line endpoints.
-
-If the dot-product (P₁ - P₀) . (P₂ - P₀) is zero (or near-zero for an
-implementation-defined definition) then gradient is ill-formed and nothing must
-be rendered.
+between which a gradient is interpolated.
 
 ![Defining points for linear gradients](images/linear_defining_points.png)
 
@@ -472,8 +467,6 @@ struct PaintLinearGradient
   VarFWORD            y0;
   VarFWORD            x1;
   VarFWORD            y1;
-  VarFWORD            x2; // Normal; Equal to (x1,y1) in simple cases.
-  VarFWORD            y2;
 };
 
 struct PaintRadialGradient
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -201,14 +201,19 @@ defining stripes in rotating colors.
...
 
-We propose definitions of linear gradients with two color line points P0 and P1
-between which a gradient is interpolated. A point P2 is defined to rotate the
-gradient angle / orientation separately from the color line endpoints.
+We propose definitions of linear gradients with two color line points P₀ and P₁
+between which a gradient is interpolated. Both points have an implicit y coordinate
+of 0. This can be adjusted via transform. A third point P₂ is implicitly present
+at (0, 1). A line through P₀, P₂ is thus initially orthogonal to a line
+through P₀, P₁. All points are subject to transform (via `PaintTransformed`).
+A shear can therefore alter the angle of the gradient.
```
drott, line 209: Why do we still need to refer to P₂? I think we no longer need to define it. The figure needs to be redrawn, but shearing or rotation is then now defined through the transformation matrix, no longer by P₂, isn't it?

colr-gradients-spec.md
```diff
@@ -468,12 +473,8 @@ struct PaintLinearGradient
 {
   uint8               format; // = 2
   Offset32<ColorLine> colorLine;
-  VarFWORD            x0;
-  VarFWORD            y0;
-  VarFWORD            x1;
-  VarFWORD            y1;
-  VarFWORD            x2; // Normal; Equal to (x1,y1) in simple cases.
-  VarFWORD            y2;
+  VarFWORD            x0;     // y0 implicitly 0
```
drott, line 476: I prefer to keep the point coordinates for `(x, y)`, as that matches the API a bit more closely in the Skia case and allows for a simple definition of radial gradients that are not rotated (assuming that's a quite common case).

In principle, since we have the color stop control, we could just define the default gradient to go from (0,0), to (0,1) and not have any points at all, but then we'd always require a transform.

```
static sk_sp<SkShader> MakeLinear(const SkPoint pts[2],
                                      const SkColor4f colors[], sk_sp<SkColorSpace> colorSpace,
                                      const SkScalar pos[], int count, SkTileMode mode,
                                      uint32_t flags, const SkMatrix* localMatrix);
```

rsheeter:
Close per f2f; not at all clear this is a good idea :)

PR #57, opened 2020-09-30
Update examples

drott:
Add example image that demonstrates circles and their coordinates
as well as repeat modes.

Preview of the change: https://github.com/drott/colr-gradients-spec/blob/radialExample/colr-gradients-spec.md#radial-gradients


rsheeter:
I like the illustration of the circles but I wonder if only coloring a triangle is going too far? Also if possible it would be nice to make it possible to edit the drawing, such as by providing it as an svg or giving the drawing program or w/e.

Continuing on to show the impact of transforms, particularly in the degenerate cases, would be really nice.

drott:
> I like the illustration of the circles but I wonder if only coloring a triangle is going too far? 

It's coloring a rectangle, but with the gradient is only defined for the triangle. I found it useful in the radial case to demonstrate for which areas the gradient is defined. Of course, making the circles larger it could fill the whole rectangle. Could you elaborate on "too far" and describe what you had in mind.  

> Also if possible it would be nice to make it possible to edit the drawing, such as by providing it as an svg or giving the drawing program or w/e.

Yes, I am making this as an SVG, but turns out Safari does not render the repeat modes other than padding correctly, so using the SVG straight in the document is not reliable. We can make an examples directory and place it there? Or put it in the examples directory of the color fonts repo?

drott:
> Continuing on to show the impact of transforms, particularly in the degenerate cases, would be really nice.

Yes, I want to illustrate the transforms, but I was thinking to do that in a separate section, separate to the gradient definitions. The SVG is prepared for that so that I can wrap the illustration in an outer transform.

behdad:
I think your rendering is wrong. Quoting the paint rule 3:

> For all values of ω where r(ω) > 0, starting with the value of ω nearest to positive infinity and ending with the value of ω nearest to negative infinity, draw the circumference of the ellipse resulting from translating circle with radius r(ω) by affine transform at position (x(ω), y(ω)), with the color at ω, but only painting on the parts of the bitmap that have not yet been painted on by earlier circles in this step for this rendering of the gradient.

You are painting for ω from +∞ to -∞, but ignoring the part "but only painting on the parts of the bitmap that have not yet been painted on by earlier circles".

Honestly every time I read this it catches me by surprise. I don't know why it was written that way (eg. front-to-back with saturation) as opposed to ω from -∞ to +∞ with no extra constraints. I'm fairly sure the two are equivalent.

drott:
Good catch, thanks for pointing that out. Phrasing this in SVG it means, to fix the rendering, I need to swap "outermost" and "focal" circle, with that, I get the rendering as expected. Outermost, is the one that maps to the 100% color stop, focal circle matches the 0% color stop.

image

which matches the one for a canvas drawn as follows:

```
var gradient = ctx.createRadialGradient(400, 200,35, 800, 200, 80);                                                                      
gradient.addColorStop(0, "#DA0025");                                                                                                     
gradient.addColorStop(0.5, "#FFC600");                                                                                                   
gradient.addColorStop(1, "#0052AB");  
```

image

which matches the algorithm we're describing.

drott:
Linear examples now added as well, illustrating P2 more clearly, adding padding modes.
Image sources contained in CL. Ready for another review.

drott:
Preview still here: https://github.com/drott/colr-gradients-spec/blob/radialExample/colr-gradients-spec.md#radial-gradients

behdad:
(There's a typo you can fix while editing. Search for "outside outside"; remove one.)

This is good. One thing not shown is color-lines where defined internal is NOT (0,1). Showing both intervals that are a subrange of that, as well as broader (eg. (-3,.5)) etc would be great as well.

drott:
> This is good. One thing not shown is color-lines where defined internal is NOT (0,1). Showing both intervals that are a subrange of that, as well as broader (eg. (-3,.5)) etc would be great as well.

Thanks for the feedback. Yes, also Rod suggested to show the simple case of P2=P0. I'll track that as an issue to update the examples further.

PR #59, opened 2020-10-01
Use Offset24 where doing so would put it alongside a single byte

### LINKED ISSUE #58: Use (new) Offset24
I'm of two views in general, both of which have not been followed in the spec consistently before:

- Use types that are wide-enough that won't run out of space, even for (reasonable but) unanticipated uses (eg. 60k axes!). Avoiding `Offset16` in most cases is part of this design philosophy.

- Don't waste bytes if not necessary. Before we've been using a lot of uint16 version numbers, even separate uint16 major/minor ones. That's a waste. Currently proposal use of `uint8_t` for `format` / `Extend` / `CompositeMode` is in line with that.

Now. The very issue of alignment that Peter seems to be obsessed with in other parts of the spec, is fully ignored here.

We should introduce an `Offset24` type, and use it next to `uint8_t` types when possible, to save a byte without limiting what's reachable, and making it faster as well because fetching one (aligned) int is faster in modern memory systems than fetching one unaligned int, which is implemented as two aligned int fetches.

rsheeter:
Fixes #58

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -467,7 +467,7 @@ struct PaintSolid
 struct PaintLinearGradient
 {
   uint8               format; // = 2
-  Offset32<ColorLine> colorLine;
+  Offset24<ColorLine> colorLine;
   VarFWORD            x0;
   VarFWORD            y0;
   VarFWORD            x1;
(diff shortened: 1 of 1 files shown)


behdad (commented):
Lgtm. Thanks.

PR #62, opened 2020-10-02
Incorporate OT/OFF style struct definitions

PeterConstable:
C++ code kept, moved into the Implementation section.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
 
 Offsets are always relative to the start of the containing struct.
 
+## Variation structures
+
+To indicate no variation, set varOuterIndex and varInnerIndex to 0xFFFF.
+
+*VarFWord record:*
+
+| Type | Name | Description |
+|-|-|-|
+| FWORD | coordinate | |
+| uint16 | varOuterIndex | |
+| uint16 | varInnerIndex | |
+
+*VarUFWord record:*
+
+| Type | Name | Description |
+|-|-|-|
+| UFWORD | distance | |
+| uint16 | varOuterIndex | |
+| uint16 | varInnerIndex | |
+
+*VarFixed record:*
+
+| Type | Name | Description |
+|-|-|-|
+| Fixed | value | |
+| uint16 | varOuterIndex | |
+| uint16 | varInnerIndex | |
+
+*Note:* In order to combine deltas with Fixed values, the ItemVariationStore format is extended to allow for int32 deltas. When combining a Fixed value with 32-bit deltas, the Fixed value is treated as though it were int32.
+
+*VarF2Dot14 record:*
+
+| Type | Name | Description |
+|-|-|-|
+| F2Dot14 | value | |
+| uint16 | varOuterIndex | |
+| uint16 | varInnerIndex | |
+
+Values are inherently limited to [-2., 2). In some contexts, limited to the [-1., 1.] or  [0., 1.].
+
+*Note:* When combining an F2Dot14 with 16-bit deltas, the F2Dot14 is treated as though it were int16.
+
+## Color structures
+
+*ColorIndex record:*
+
+| Type | Name | Description |
+|-|-|-|
+| uint16 | paletteIndex | Index for a CPAL palette entry. |
+| VarF2Dot14 | alpha | Variable alpha value. |
+
+The alpha.value is limited to [0., 1.].
+
+*ColorStop record:*
+
+| Type | Name | Description |
+|-|-|-|
+| VarF2Dot14 | stopOffset | Proportional distance on a color line; variable. |
+| ColorIndex | color | |
+
+The stopOffset.value is limited to [0., 1.].
+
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
 
 Offsets are always relative to the start of the containing struct.
 
+## Variation structures
+
+To indicate no variation, set varOuterIndex and varInnerIndex to 0xFFFF.
+
```
rsheeter, line 352: Suggest retention of the groupings, e.g. `// Base types` could become `### Base types`

colr-gradients-spec.md
```diff
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
...
+| uint16 | varOuterIndex | |
+| uint16 | varInnerIndex | |
+
+Values are inherently limited to [-2., 2). In some contexts, limited to the [-1., 1.] or  [0., 1.].
+
+*Note:* When combining an F2Dot14 with 16-bit deltas, the F2Dot14 is treated as though it were int16.
+
+## Color structures
+
+*ColorIndex record:*
```
rsheeter, line 393: Suggestion retention of 
```
// The ColorIndex alpha is multiplied into the CPAL color looked up using
// paletteIndex to produce a final color
```

(reformatted to new layout ofc)

colr-gradients-spec.md
PeterConstable: Do you mean in the C++ code (implying that the one md code block becomes multiple code blocks)?

colr-gradients-spec.md
```diff
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
...
+
+| Type | Name | Description |
+|-|-|-|
+| uint16 | paletteIndex | Index for a CPAL palette entry. |
+| VarF2Dot14 | alpha | Variable alpha value. |
```
rsheeter, line 398: Description: Default 1.0. Values outside [0.,1.] reserved.

You can then delete `The alpha.value is limited to [0., 1.].` The default and the notion the values are reserved is, I think, significant.

PeterConstable: VarF2Dot14 might later get used in other contexts in which range isn't limited to [0,1]. I think it's more appropriate to state restrictions on range to the context in which it's used.

colr-gradients-spec.md
```diff
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
...
+| 21 | COMPOSITE_EXCLUSION | See [exclusion blend mode][23] |
+| 22 | COMPOSITE_MULTIPLY | See [multiply blend mode][24] |
+| | *Non-separable color blend modes:* | |
+| 23 | COMPOSITE_HSL_HUE | See [hue blend mode][25] |
+| 24 | COMPOSITE_HSL_SATURATION | See [saturation blend mode][26] |
+| 25 | COMPOSITE_HSL_COLOR | See [color blend mode][27] |
+| 26 | COMPOSITE_HSL_LUMINOSITY | See [luminosity blend mode][28] |
+
+*Affine2x3 record:*
+
```
rsheeter, line 469: Perhaps a note that this is a std 2x3 transformation matrix?

colr-gradients-spec.md
```diff
@@ -346,6 +346,252 @@ the COLR v1 glyph structure.
...
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 2. |
+| Offset24 | colorLineOffset | Offset to ColorLine, from start of PaintLinearGradient table. |
+| VarFWord | x0 | Start point x coordinate. |
+| VarFWord | y0 | Start point y coordinate. |
+| VarFWord | x1 | End point x coordinate. |
+| VarFWord | y1 | End point y coordinate. |
+| VarFWord | x2 | Rotation vector end point x coordinate. |
```
rsheeter, line 498: I think we should keep the note that x2, y2 are equal to x1, y1 in simple cases here somewhere

colr-gradients-spec.md
rsheeter: I mean translate the comments from C++ that indicate groups of structs into md headings. For example:

```
## Variation structures

### Base types
^ the md equivalent of // Base types in the C++ version

<...the base types...>

## Building blocks

...etc...
```

colr-gradients-spec.md
rsheeter: I'm suggesting the change be made to the description of the field ColorIndex.alpha, not on the general definition of VarF2Dot14.

PeterConstable: In what sense is 1.0 a default wrt the structure definition? It may be a default used in an implementation creating a new instance in memory; but when parsing font data the value is whatever is given in the data.

colr-gradients-spec.md
PeterConstable: In that case, I think it may make more sense to re-org a bit the C++ (which already has some things out of logical order anyway).

PR #63, opened 2020-10-02
Create CODE_OF_CONDUCT.md

davelab6:


behdad:
Is truly bizarre how people are obsessed with Code of Conduct, but there's nothing to stop people like Peter & Vlad from what is clearly called administrative abuse.

behdad:
Constantly questioning every little aspect of a design, that took a year to come up with, and demand explanation IS ABUSE.

drott (approved):

Issue #60, opened 2020-10-02
Clarify PaintFilledGlyph


rsheeter:
It's clearly misleading; it clips not fills. `PaintClipGlyph`?

Explain nesting of clips == intersection

rsheeter:
Helps #54

behdad:
It fills the glyph outline with paint. I don't see any problem.

True bikeshedding. That's why we should have just kept it as PaintGlyph. Sigh

PeterConstable:
Suggestion: PaintGlyphOutlineClip

behdad:
Just argh...

PeterConstable:
Didn't notice `PaintClipGlyph`. That works.

behdad:
`PaintClipGlyph` is the most bizarre naming.

Please remove my name from this proposal. This is not what I envisioned anymore.

alerque:
Reading the before and after from 6d46c32 along with descriptions of each I have to agree with @behdad here that the original pair of `PaintGlyph` and `PaintColrGlyph` was clearer and more to the point. The confusion pointed out in this issue as originally opened was only brought on by the change from `PaintGlyph`→`PaintFilledGlyph`, and I think `PaintClipGlyph` just muddies the waters more.

rsheeter:
Ignoring the name for the moment, I think there is a fair criticism that how things combine was not clear. If you think of PaintGlyph (by any name :) as a clip then the result of PaintGlyph nesting another PaintGlyph is obvious. If you think of it as a fill ... possibly less clear.

> The confusion pointed out in this issue as originally opened was only brought on by the change from PaintGlyph→PaintFilledGlyph, and I think PaintClipGlyph just muddies the waters more.

I think I agree and that I should just reverse that change

behdad:
> Ignoring the name for the moment, I think there is a fair criticism that how things combine was not clear.

That's not the point of a file format specification though. You or Peter can write blogposts about how things are expected to work together. The font format is not room for Peter's ruminations of his limited understanding of how things are supposed to work.

cc @simoncozens @alerque

rsheeter:
My inclination at this point is to go back to the name `PaintGlyph` and add (or restore, I think I actually had it at one point) a note that what it does is clip the paint it references. 

> That's not the point of a file format specification though

There is value in some degree of explanation of how to use the format. That explanation should fit the audience, we can err in providing too much or too little detail.

[...]

PR #65, opened 2020-10-05
Rename PaintFilledGlyph to PaintClipGlyph

drott:
`PaintClipGlyph` more clearly describes what graphics context operation is
executed before nested paints are executed.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -88,7 +88,7 @@ are defined:
 1. **Solid** paints a solid color
 1. **Linear gradient** paints a linear gradient
 1. **Radial gradient** paints a radial gradient
-1. **Filled Glyph** draws a non-COLR glyph
+1. **Clip Glyph** draws a non-COLR glyph
    * A COLR v1 glyph made up of a list of Glyph paints is equivalent to a COLR v0 Layer Record with the added ability to use gradients.
 1. **COLR Glyph** reuses a COLR v1 glyph at a new position in the graph
 1. **Transformed** reuses another paint, applying an affine transformation
@@ -304,7 +304,7 @@ not render.
 
 The following paints are always bounded:
 
-- `PaintFilledGlyph`
+- `PaintClipGlyph`
 - `PaintColrGlyph`
 
 The following paints are always unbounded:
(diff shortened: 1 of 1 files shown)


drott:
Closing for now, considering discussion in #60

behdad:
This is bizarre. One implementation may use fill operation while another uses clipping. Just argh to bending to Peter's unreasonable demands.

drott:
I don't understand how it would be a fill operation in all scenarios. Could you explain how you'd see a fill operation used generically in all cases? If we `PaintFilledGlyph` followed by `PaintSolid|RadialGradient|LinearGradient` it can be a fill operation. That's how I had it implemented for our initial draft and I agree that for these cases `PaintFilledGlyph` is reasonable. But if nested with `PaintComposite`, `PaintTransform` and `PaintColorGlyph`, isn't it more generic to clip the graphics context, then do an unbounded paint of the respective gradient/solid or recursively execute the other operations? That's why I think the name would make sense.

behdad:
```
  cairo_push_group (cr);
  process (cr, paint);
  cairo_pop_group_to_source (cr);
  cairo_show_glyphs (cr, gid);
```
whereas `cairo_show_glyphs()` is equivalent to `cairo_glyph_path(); cairo_fill()` but going through the glyph-caching subsystem and the font rasterizer instead of the generic fill.

The point is, HOW this is implemented is an implementation detail of particular implementations. Not all graphics systems even have a clip operation that works on arbitrary shapes!

The point is, and always was, that this operation draws a glyph, painted with some paint. That's all. `PaintGlyph`. Just because someone came along and said "I object" doesn't mean it had to be changed.

alerque:
> I don't understand how it would be a fill operation in all scenarios.

I don't think that's what is being argued. See #60, but Behdad's point –that I agreed with– was that it shouldn't have been renamed from `PaintGlyph` to `PaintFilledGlyph` in the first place, and `PaintClipGlyph` doesn't make sense either. The latter two more 'specific' names both seem to dictate a specific way of implementing it whereas the original `PaintGlyph` just roughly describes the outcome differentiates it from `PaintColrGlyph`.

PR #66, opened 2020-10-05
Fix ColorIndex in PaintSolid

drott:


anthrotype (approved):
LGTM

we forgot to update this when we changed the name of Color struct type to ColorIndex

PR #71, opened 2020-10-05
Stop offset is not limited to 0..1

rsheeter:


anthrotype:
There's still another mention of "Values outside [0.,1.] reserved" left in the "C++ Structures" section, under `ColorStop` struct

PR #76, opened 2020-10-06
Remove comment regarding reserved interval

drott:


drott:
CC @PeterConstable, fyi - it's not excluded to specify the color stops on interval stops outside the range, although we should clarify the details further.

PeterConstable:
Looks like your change is allowing alpha outside the range [0,1], not color stop index. Presumably alpha should be limited (and clipped if a variation with deltas applied goes outside the range).

drott:
> Looks like your change is allowing alpha outside the range [0,1], not color stop index. Presumably alpha should be limited (and clipped if a variation with deltas applied goes outside the range).

You're right. Then it seems that this already had been removed in the structs section, updated the change to remove the respective comment from the section containing the structs.

PeterConstable (approved):

PeterConstable:
So, how are extend mode behaviours affected by colour stops outside the [0,1] range? 

Are the extend behaviours defined as extending from [0,1] positions, or from the first and last color stop?

For PAD, this difference will affect what colours are used in a range [1, N] if a colour stop is set for N >1. (Similarly for the "start" end of the line.)

For REPEAT and MIRROR, this difference affects what range is repeated. For example, suppose colour stops are defined at 0 and 0.75. Does the first repeat/mirror (on the "end" end of the line) go from 0.75 to 1.5, or from 1.0 to 2.0?

Or if the stops are defined at 0 and at 1.1, does the first repeat/mirror go from 1.1 to 2.2? From 1.0 to 2.0 (superseding the original colours from 1.0 to 1.1)?

rsheeter:
My understanding is that extend is defined based on [0,1]. A stop at say 1.1 doesn't change that, it just influences how you compute what goes at 1.0.

PeterConstable:
If stops are defined at 0 and 1.1, and the color value for 1.0 is interpolated (as it must), might that imply that, for a color line defined by stops at 0 and 0.75, then the color value for 1.0 is similarly *extrapolated*?

I'm guessing the intent is that values from 0.75 to 1.0 would be padded, regardless of the extend mode? Whichever it is, these details need to be clear.

drott:
The extend modes pad, repeat, reflect based on the defined interval. 0, and 1 are intended as landmarks for where you can place color stops if you want them anchored at the particular gradient coordinates.

> For REPEAT and MIRROR, this difference affects what range is repeated. For example, suppose colour stops are defined at 0 and 0.75. Does the first repeat/mirror (on the "end" end of the line) go from 0.75 to 1.5, or from 1.0 to 2.0?

From 0.75 to 1.5.

> Or if the stops are defined at 0 and at 1.1, does the first repeat/mirror go from 1.1 to 2.2? From 1.0 to 2.0 (superseding the original colours from 1.0 to 1.1)?

From 1.1 to 2.2.

And so on, if you have color stops from -5 to 5, then the first repeat goes from 5 to 15, and on the other side from -5 to -15.

drott:
Tracked in #77

PR #85, opened 2020-10-22
Matrix order explanation

drott:


anthrotype (approved):

behdad:
This looks wrong. The vastly common convention is that `xy` component is the coefficient of input `y` in output `x`. That is:
```
 x' = xx * x + xy * y + dx
```
Your matrix formulation seems to get it otherwise.

anthrotype:
2D Affine transformation matrix is written as

```
| scaleX | skewX  | translateX |
| skewY  | scaleY | translateY |
|      0 |      0 |          1 |
``` 

In SkMatrix API, this is stored as an array of 9 scalars in _row-major order_, i.e.

`[scaleX, skewX, translateX, skewY, scaleY, translateY, 0, 0, 1]`

whereas in COLRv1 Affine2x3 (as well as in FT_Matrix, or FontTools Transform class, SVG transforms, PDF, etc.), the same matrix is stored as a sequence of 6 scalars ordered in _column-major order_ (with the last three "augmented vector" omitted):

`[scaleX, skewY, skewX, scaleY, translateX, translateY]` ==> `[xx, xy, yx, yy, dx, dy]`

see https://github.com/google/skia/blob/b432d0b757655ae0f855a4f138b196d2da9ee8a1/include/core/SkMatrix.h#L313-L334

So, `xy` corresponds to `skewY` and `yx` corresponds to `skewX`.

Am I completely wrong?

rsheeter:
I interpret [xx xy] as î, [yx yy] as ĵ and ... seemingly end up with @drott formulation of

|  *î*  |  *ĵ* | |
| --- | --- | --- |
| xx  | yx | dy |
| xy | yy  | dx |

If that's correct I suggest that instead of providing the example table we might instead add descriptions to the fields (currently first four have no description):

| Type | Name | Description |
|-|-|-|
| VarFixed | xx | **x part of x-basis vector** |
| VarFixed | xy | **y part of x-basis vector** |
| VarFixed | yx | **x part of y-basis vector** |
| VarFixed | yy | **y part of y-basis vector** |
| VarFixed | dx | Translation in x direction. |

rsheeter:
https://github.com/vishnubob/cairo/blob/master/src/cairo.h#L174

anthrotype:
I'm still confused.. In terms of `skewX` and `skewY, to use the terminology from [SkMatrix API](https://github.com/google/skia/blob/b432d0b757655ae0f855a4f138b196d2da9ee8a1/include/core/SkMatrix.h#L313-L334), what do your `xy` and `yx` correspond to?

AFAIU, this is the following:

xx == scaleX (horizontal scale factor)
xy == skewY (vertical skew factor)
yx == skewX (horizontal skew factor)
yy == scaleY (vertical scale factor)
dx == translateX (horizontal translation)
dy == translateY (vertical translation)

FontTools Transform class uses the same (xx, xy, yx, yy, dx, dy) terminology used here for Affine2x3, and the method that maps a (x, y) point by a Transform is defined like this:

```python
	def transformPoint(self, p):
		"""Transform a point.
		Example:
			>>> t = Transform()
			>>> t = t.scale(2.5, 5.5)
			>>> t.transformPoint((100, 100))
			(250.0, 550.0)
		"""
		(x, y) = p
		xx, xy, yx, yy, dx, dy = self
		return (xx*x + yx*y + dx, xy*x + yy*y + dy)
```

https://github.com/fonttools/fonttools/blob/28a215173a2ddd0d3261607f4af5f4d64c1d3055/Lib/fontTools/misc/transform.py#L159-L160

Has that been wrong all along?

anthrotype:
Ignoring for a second the xy/yx terminology -- take the more neutral picosvg's `Affine2D(a, b, c, d, e, f)` class.
There we have:

```python
    def map_point(self, pt: Tuple[float, float]) -> Point:
        x, y = pt
        return Point(self.a * x + self.c * y + self.e, self.b * x + self.d * y + self.f)
```

https://github.com/googlefonts/picosvg/blob/09d05fe22ded144bd53af6e58133ce8b146c0588/src/picosvg/svg_transform.py#L138-L141

What is `b` and `c` in this case?

justvanrossum:
If I compare `fontTools.misc.transform.Transform.transformPoint()` with `map_point()` above, I see the same order of the fields being applied in the same way:

```python
	def transformPoint(self, p):
		(x, y) = p
		xx, xy, yx, yy, dx, dy = self
		return (xx*x + yx*y + dx, xy*x + yy*y + dy)
```
Ie. `c` maps to `yx` and `d` maps to `xy`.

https://github.com/fonttools/fonttools/blob/28a215173a2ddd0d3261607f4af5f4d64c1d3055/Lib/fontTools/misc/transform.py#L149-L160

justvanrossum:
Also: I pass fonttools Transform objects directly (well, via skia-python) to Skia's `SkMatrix::setAffine` and that works as expected as far as I can tell.

justvanrossum:
The fonttools Transform value order also matches the order of PostScript transformations.

PostScript:
```postscript
[1 0.176327 0 1 0 0] concat

newpath
100 100 moveto
0 300 rlineto
200 0 rlineto
0 -300 rlineto
-200 0 rlineto
closepath
fill
showpage
```

DrawBot:
```python
import math
from fontTools.misc.transform import Transform

t = Transform()
t = t.skew(0, math.radians(10))
print(t)  # <Transform [1 0.176327 0 1 0 0]>
transform(t)

rect(100, 100, 200, 300)
```

rsheeter:
I'm inclined to update our Affine2D to something like:

```
x_basis: Vector
y_basis: Vector
offset: Vector
```

I think map_point will be dramatically clearer than a mess of abc... or xx,xy,etc.

justvanrossum:
But the Skia matrix is a 3x3, and indeed the mapping between Affine fields and Skia's 3x3 fields is not super intuitive:
```python
>>> import skia
>>> import math
>>> from fontTools.misc.transform import Transform
>>> t = Transform()
>>> t = t.skew(0, math.radians(10))
>>> t
<Transform [1 0.176327 0 1 0 0]>
>>> m = skia.Matrix()
>>> m.setAffine(t)
<skia.Matrix object at 0x7fb8ee98ecb0>
>>> list(m)
[1.0, 0.0, 0.0, 0.17632697522640228, 1.0, 0.0, 0.0, 0.0, 1.0]
>>> m.mapPoints([(100, 100)])
[Point(100, 117.633)]
>>> t.transformPoints([(100, 100)])
[(100.0, 117.6326980708465)]
>>> 
```

justvanrossum:
FWIW, the abcdef notation matches this explanation: https://scientificgems.wordpress.com/2014/11/28/mathematics-in-action-affine-transformations-and-postscript/
image

justvanrossum:
> x_basis: Vector
> y_basis: Vector
> offset: Vector

Decomposing what's essentially a 2x2 matrix plus an x, y offset into three vectors does not make sense to me. It is very different from all explanations I've seen so far, and I doubt it will be helpful for anyone who's ever dealt with Affine transformations, or matrices in general.

rsheeter:
My concern with abcd or xx/xy/etc is that these don't explain what the fields mean / why it is the way it is and we're observing this confuse people both on this thread and in the code for COLR emoji processing. There must be something we can do to make it a little less magical.

https://scientificgems.wordpress.com/2014/11/28/mathematics-in-action-affine-transformations-and-postscript/ similarly doesn't really explain why, it just asserts: "They are defined by a combination of a matrix multiplication and a translation".

I quite like how https://en.wikipedia.org/wiki/Affine_transformation#/media/File:2D_affine_transformation_matrix.svg depicts it: blue is the x-basis, red is the y-basis, and suddenly the magic matrix seems a little less magic. Why do 90 degree rotations look like that? - makes perfect sense viewed in terms of basis vectors.

anthrotype:
> Decomposing what's essentially a 2x2 matrix plus an x, y offset into three vectors does not make sense to me. It is very different from all explanations I've seen so far

I suggest you watch 3Blue1Brown's [video](https://youtu.be/kYB8IZa5AuE) on YouTube where he presents linear transformations as linear combinations of the two basis or unit vectors, `î = (1, 0)` for horizontal x axis and `ĵ = (0, 1)` for vertical y axis -- actually all 3Blue1Brown's linear algebra videos are worth watching (thanks Rod for sharing!). 
This definition is much clearer and more intutitive than those based on memorizing rules for matrix multiplication.

justvanrossum:
> I suggest you watch 3Blue1Brown's video 

Thanks, that was very good indeed!

drott:
Updated and hopefully matching what we discussed in f2f and what @rsheeter outlined in https://github.com/googlefonts/colr-gradients-spec/pull/85#issuecomment-718160362. 

> I'm inclined to update our Affine2D to something like:
> 
> ```
> x_basis: Vector
> y_basis: Vector
> offset: Vector
> ```

Agree with Just in https://github.com/googlefonts/colr-gradients-spec/pull/85#issuecomment-718975556  in preferring to keep talking about a matrix, in particular as this keeps things consistent when talking about how nested transforms are defined and how matrices are concatenated.

anthrotype:
I agree with the @rsheeter's conclusions from https://github.com/googlefonts/colr-gradients-spec/pull/85#issuecomment-714891901

However there's still a discrepancy as regards the meaning/order of `xy` vs `yx` between FontTools' Transform class on the one hand, and Cairo and FreeType matrix struct definitions.

Given the following (identity) affine matrix, with the columns containing respectively the x basis vector `î`, the y basis vector `ĵ` and the translation vector; and the rows containing their horizontal (x) and vertical (y) coordinates:

|   |  *î*  |  *ĵ* |   |
| --- | --- | --- | --- |
| *x* | 1  | 0 |  0 |
| *y* | 0 | 1  | 0 |

In FontTools Transform, the field named `yx` indicates `ĵ.x`, i.e. the x part of the y basis vector, second column, first row; whereas `xy` means `î.y`, i.e. the y part of the x basis vector, first column, second row.
This is evident from the way these fields are used in the `transformPoint` method quoted earlier, in particular this line:

```python
		return (xx*x + yx*y + dx, xy*x + yy*y + dy)
```

Thus in FontTools, the affine 2x3 matrix looks like this:

|   |  *î*  |  *ĵ* |  |
| --- | --- | --- | --- |
| *x* | xx  | **yx** |  dx |
| *y* | **xy** | yy  | dy |

The 6 values are stored in a tuple in the following order: (xx, xy, yx, yy, dx, dy). The matrix is stored in column-major order (columns first, then rows).

***

Now, [Cairo](https://github.com/vishnubob/cairo/blob/cffa452f44eadebef8553502e3d6cc49829d38ce/src/cairo.h#L174-L196) and [FreeType](https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_matrix) use the same terminology as FontTools (xy, yx, etc.) but assign it a different meaning.

[`FT_Matrix`](https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_matrix) (2x2 matrix) is defined as 4 Fixed numbers: xx, xy, yx, yy

```c
  typedef struct  FT_Matrix_
  {
    FT_Fixed  xx, xy;
    FT_Fixed  yx, yy;

  } FT_Matrix;
```

By just looking at the labels one could assume that they mean the same as in FontTools, however the ft header explains what the values actually mean (note how `xy` and `yx` are used in opposite places than in the FontTools snippet above):

```
x' = x*xx + y*xy
y' = x*yx + y*yy
```

So unlike in FontTools, in FT_Matrix `xy` is used to indicate `ĵ.x`, i.e. the x part of the y basis vector, second column, first row in the table; whereas `yx` means `î.y`, i.e. the y part of the x basis vector, first column, second row.
FT_Matrix serializes the matrix values in _row-major_ order, rows first, then columns.

As for [Cairo matrix struct](https://github.com/vishnubob/cairo/blob/cffa452f44eadebef8553502e3d6cc49829d38ce/src/cairo.h#L174-L196), the 2x3 matrix for affine transformations is defined as a struct containing 6 doubles in this order: xx, yx, xy, yy, x0, y0.

The meaning of these is clarified by the formula in the Cairo header comments:

```
x_new = xx * x + xy * y + x0;
y_new = yx * x + yy * y + y0;
```

Like in FreeType but unlike FontTools, also in Cairo the `xy` value means `ĵ.x` (x part of the y basis vector, second column, first row), and `yx` means `î.y` (y part of the x basis vector, first column, second row).
Unlike FreeType and similarly to FontTools, Cairo matrix is stored in _column-major_ order.

In our current COLRv1 draft we opted for the same labels in Affine2x3 definition: xx, xy, yx, yy, dx, dy.

As I tried to explain by contrasting FontTools and FreeType/Cairo, this terminology is tarnished with confusion (xy and yx mean different things depending who you ask) making things more obscure than they should be.

So here's my proposal.

I think we should avoid the use of xx, xy, etc. labels altogether and make explicit references to the basis vectors in the Affine2x3 fields, like so:

| VarFixed | ix | x part of î basis vector |
| VarFixed | iy | y part of î basis vector |
| VarFixed | jx | x part of ĵ basis vector |
| VarFixed | jy | y part of ĵ basis vector |
| VarFixed | dx | x part of translation vector |
| VarFixed | dy | y part of translation vector |

Also, I would like to advocate for a _column-major_ ordering of the matrix fields when serialized to a list of floats.
In Dominik's current draft, the order is _row-major_ for the first 4 fields (i.e. the 2x2 part of the 2x3 matrix); whereas the the last two dx/dy are in column-major order, which I find a bit contradicting.
Column-major order is the one used by FontTools, Cairo, SkMatrix::setAffine, PDF and SVG matrices.
Alternatively if you prefer to go for row-major ordering, I suggest we reshuffle the fields so they are consistently all in row-major order.
That is: ix, jx, dx, jx, jy, dy. Skia's SkMatrix internally uses row-major order, for example.
I'd still prefer column-major order so that our current compiler will continue to work unmodified.

colr-gradients-spec.md
```diff
@@ -590,14 +590,22 @@ Supported composition modes are taken from the W3C [Compositing and Blending Lev
...
+| VarFixed | yy | y part of y-basis vector |
 | VarFixed | dx | Translation in x direction. |
 | VarFixed | dy | Translation in y direction. |
 
-This is a standard 2x3 matrix for 2D affine transformation.
+The `Affine2x3` record is a 2x3 matrix for 2D affine transformations, so
+that for a transformation matrix _M_ and an existing vector _v = (x, y)_
+the mapped vector _v'_ is calculated as
+
+_v' = M * v = (xx * x + xy * y + dx, xy * x + yy * y + dy)_
```
anthrotype, line 604: look there's a typo (not by chance!)

```suggestion
_v' = M * v = (xx * x + xy * y + dx, yx * x + yy * y + dy)_
```

justvanrossum:
(The xy vs yx naming confusion/debate doesn't seem to be unique: although Cairo and FreeType seem to agree with each other, a quick search yields several examples that agree with fonttools. So: I agree that it's probably better not to use that naming scheme. But I object to fonttools being "wrong" :) Also: these names were private until fontTools' Transform class became a NamedTuple.)

drott:
Sorry for introducing additional confusion with the typo. I updated the proposed new wording. 

I prefer not to introduce a third naming convention (ix, jx, etc.) and I would prefer to actually not talk about basis vectors in giving the fields names, as the matrix does not contain basis vectors, but coefficients. I am okay with mentioning that in the Note, see new draft.

> Also, I would like to advocate for a _column-major_ ordering of the matrix fields when serialized to a list of floats.
> In Dominik's current draft, the order is _row-major_ for the first 4 fields (i.e. the 2x2 part of the 2x3 matrix); whereas the the last two dx/dy are in column-major order, which I find a bit contradicting.

I've taken into account the inconsistency that Cosimo pointed out with regards to column-major order vs. row-major order.

colr-gradients-spec.md
drott: Hopefully fixed.

drott:
> I would prefer to actually not talk about basis vectors in giving the fields names

I would prefer not to, but I am okay with it, and since Cosimo said in chat, that the "coefficient of" language that I used above is not necessarily better to understand, I put the "... part of ... basis: vector format back.

justvanrossum:
I bet the Cairo/FreeType variant has a good explanation, too, but I find fonttools more logical. I like [this explanation](http://cdn.sencha.com/touch/sencha-touch-2.1.0/docs/#!/api/Ext.draw.Matrix):
```
xx: Coefficient from x to x
xy: Coefficient from x to y
yx: Coefficient from y to x
yy: Coefficient from y to y
dx: Offset of x
dy: Offset of y
```
...which I find quite elegant, and matches the fonttools names.

I also like the fact that the first letter in the name changes with the horizontal position in the matrix, and the second letter changes with the vertical:
```
| xx yx |
| xy yy |
```

anthrotype:
I'm pretty deep in this bikeshed, so be patient with me.

How about we simply say: the COLRv1 Affine2x3 matrix values are listed in column-major order, and then we distinguish the fields using zero-based _indices into columns then rows_: e.g. `scale00`, `scale01`, `scale10`, `scale11`, `translateX`, `translateY`.
Technically these are "scalar" values, so "scale" would make sense as a common prefix. Or even just `transform00`, `transform01` etc. for the first 2x2 linear transformation + the translation vector.

I got the idea from the [OpenType `glyf` specification](https://docs.microsoft.com/en-us/typography/opentype/spec/glyf), in which components can define a 2x2 linear transformation when `WE_HAVE_A_TWO_BY_TWO` flag is present. In the pseudocode, the incriminated matrix components are named `scale01` and `scale10`:

```c
    } else if ( flags & WE_HAVE_A_TWO_BY_TWO ) {
        F2Dot14  xscale;    /* Format 2.14 */
        F2Dot14  scale01;   /* Format 2.14 */
        F2Dot14  scale10;   /* Format 2.14 */
        F2Dot14  yscale;    /* Format 2.14 */
    }
```

Of course the OpenType spec doesn't say what these `0` and `1` in the `scale01` and `scale10` names actually mean.
I had to look at the ttgload.c source file in FreeType to understand that they are listed in column-major order.
You can tell because when freetype parses them into its own `FT_Matrix` struct, it **flips** the order of `yx` and `xy` (see.[`ttgload.c`](https://github.com/drott/freetype2-colr/blob/80389cab7f5823e7c3b3fe4190a7c337439317de/src/truetype/ttgload.c#L686-L697) source file) -- and we have ascertained from my earlier discussion that `FT_Matrix` struct stores the 2x2 values in _row-major_ order.

rsheeter:
@drott current draft LGTM

rsheeter:
IIUC (courtesy of @anthrotype) this matches Cairo name and order, FreeType name but not order, and FontTools order but not name. We are at least consistent with some things and between formula and descriptions I think interpretation is relatively clear.

behdad:
I disagree with the field swap, but whatever... I might record a one-hour video to explain why the xx,xy,yx,yy is used and is superior.  Anyway...

anthrotype:
@behdad are you suggesting we go for a _row-major_ order (like FT_Matrix 2x2, and like SkMatrix) instead of a _column-major_ order (like Cairo matrix, FontTools Transform)?

If we do so, I propose we place the translation dx,dy _not_ at the end after the xx,xy,yx,yy -- which would mix a row-major order for the 2x2 with a column order for the last column dx,dy. But rather we list them all 6 in row-major order: that is: xx, xy, dx, yx, yy, dy. That is how SkMatrix stores them internally.

rsheeter:
I'm willing to take it on faith that @behdad is right about advantage of order so I suggest we just don't flip the fields and call it a day.

justvanrossum:
If you use a sequence of 6 numbers to describe an Affine transformation, it would be very odd not to use the order that, like, _everyone_ is using. If it's about the names for those 6 numbers, well, then I'm out of opinions :)

behdad:
If every API has it the same, then sure do it the same. I was under the impression that both are done in the wild, but might be wrong.

behdad:
In particular, I thought what I'm suggesting matches cairo. But I just checked and I was wrong.

I remember there was a very lengthy discussion about this when cairo API was being designed back in ~2005 as well. :))

rsheeter:
Here are the fields and meanings (in terms of basis vectors) I seem to see across a few implementations; I have used tx, ty to denote translation in the basis part rows as I'm not sure about a conventional name for this.

| Struct | | | | | | |
| --- | --- | --- | --- | --- | --- | --- |
| [ft_matrix](https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_matrix) fields | xx | xy | yx | yy | | |
| [ft_matrix](https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_matrix) basis parts | î.x | ĵ.x | î.y | ĵ.y | 
| [_cairo_matrix](https://github.com/vishnubob/cairo/blob/master/src/cairo.h#L174) fields | xx | yx | xy | yy | x0 | y0 |
| [_cairo_matrix](https://github.com/vishnubob/cairo/blob/master/src/cairo.h#L174) basis parts | î.x | î.y | ĵ.x | ĵ.y | tx | ty |
| [transform.py](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/misc/transform.py#L149) fields | xx | xy | yx | yy | dx | dy |
| [transform.py](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/misc/transform.py#L149) | î.x | î.y | ĵ.x | ĵ.y | tx | ty |
| [SkMatrix](https://github.com/google/skia/blob/master/include/core/SkMatrix.h) internal fields* | ScaleX | SkewX | TransX | SkewY | ScaleY | TransY |
| [SkMatrix](https://github.com/google/skia/blob/master/include/core/SkMatrix.h) internal |  î.x | ĵ.x| tx | î.y | ĵ.y | ty |
| [SkMatrix](https://github.com/google/skia/blob/master/include/core/SkMatrix.h) asAffine** fields | ScaleX | SkewY | SkewX | ScaleY | TransX | TransY |
| [SkMatrix](https://github.com/google/skia/blob/master/include/core/SkMatrix.h) asAffine basis parts | î.x | î.y | ĵ.x | ĵ.y | tx | ty |

_*_ SkMatrix fields are named based on [consts](https://github.com/google/skia/blob/2882e70ef377798b17f1b2d7bd5d8d4c4582d246/include/core/SkMatrix.h#L316) that offset into the underlying array
_**_ asAffine matches PDF and XPS order according to comments

Let's at least avoid introducing a *new* order or interpretation of names lol

rsheeter:
I suggest we might merge what @drott currently has: î.x, î.y, ĵ.x, ĵ.y, tx, ty using the names xx, yx, xy, yy, dx, dy. That is, the x-basis, then the y-basis, then the translation. IIUC if we compare field meaning (ignore names) this matches _cairo_matrix, FontTools, PDF and XPS, and SkMatrix::asAffine.

anthrotype:
Sgtm

rsheeter:
Next time we get to 100 comments!

PR #87, opened 2020-10-26
Allow reuse of a subset of the layers of another COLR glyph

### LINKED ISSUE #86: Adjust PaintColrGlyph to directly express subset of layers
In playing with PaintColrGlyph for Noto Emoji I realized that to reuse a subset of layers from a glyph you end up having to extract a new BaseGlyphV1Record and point all users at it. This makes reuse of a small # of layers result in an increase in filesize.

Visualization of what the layer graphs currently end up looking like @ https://codepen.io/rs42/pen/qBNRwew?editors=1000.

Thinking "aloud", what if we amend format 5 to directly express of "reuse some layers"? - sketch:

```C++
// 5 bytes. 
// In a LayerV1List each pointer costs 4 bytes
// For a single paint you just point directly, for 2+ consecutive you "win" by using format 5
PaintColrSlice
  uint8 format // 5
  uint16 glyphID
  uint8 firstLayer
  uint8 lastLayer
```

In this model the first clock face happily draws it's layers, the second comes along and wherever it has a run of at least 2 of the same layers it sticks in a format 5 entry.

As @anthrotype noted when I initially filed this on the wrong repo, this also saves consumption of a glyphid when you extract a set of layers into a new base record so they can be shared.

In local test the Noto clock faces (mostly reused shapes) compiled to a 3308 font without PaintColrGlyph, a 3248 byte font with the old PaintColrGlyph and a 2436 byte font with the propose PaintColrSlice.

rsheeter:
Allow reuse of a slice of another COLR glyph

Fixes #86 

EDIT: generalized to a new paint format that points to a series of paint pointers, ty @behdad

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -98,7 +98,7 @@ Several different types of paint are defined:
    compositing rule to combine them
    * We draw on https://www.w3.org/TR/compositing-1/ for our composite modes
 1. **[COLR Glyph](#colr-glyph)** reuses a COLR v1 glyph at a new position in the graph
-1. **[COLR Layers](#colr-layers)** reuses one or more layers
+1. **[COLR Layers](#colr-layers)** paint one or more layers
 
 We have added an "alpha" (transparency) scalar to each invocation of a palette
 color. This allows for the expression of translucent versions of palette
@@ -316,12 +316,14 @@ glyph id `gid` specified in `PaintColrGlyph`. See section [Reusable Parts](#reus
 
 ## COLR Layers
 
-`PaintColrLayers` is a special paint which allows reuse of some or all of a COLR
-v1 glyph as a paint. It points to a sequence of Paint pointers and specifies how many
-to consume. This allows reuse of partial COLR Glyphs. For example, suppose glyph
+`PaintColrLayers` points to a sequence of Paint pointers and specifies how many
+to consume. The mechanism can be used to share layers or sequences of layers
+between multiple COLR glyphs.
+
+*Example:* Suppose glyph
 A has 10 layers, 3 of which are a common backdrop that glyph B also uses. Glyph B
-can define a PaintColrLayers record that points into A's layers and specifies how
-many to take.
+can define a PaintColrLayers record that points to the same layers as A for the
+common parts.
 
 See section [Reusable Parts](#reusable-parts).
 
@@ -369,6 +371,7 @@ Offsets are always relative to the start of the containing struct.
 | Offset32 | layerRecordsOffset | Offset to layerRecords array (may be NULL). |
 | uint16 | numLayerRecords | May be 0 in a version 1 table. |
 | Offset32 | baseGlyphV1ListOffset | Offset to BaseGlyphV1List table. |
+| Offset32 | layersV1Offset | Offset to LayerV1List table. |
 | Offset32 | itemVariationStoreOffset | Offset to ItemVariationStore (may be NULL). |
 
 ##### BaseGlyphV1List table
@@ -391,10 +394,11 @@ Offsets are always relative to the start of the containing struct.
 
 | Type | Field name | Description |
 |-|-|-|
-| uint8 | numLayers |  |
-| Offset32 | paintOffset[numLayers] | Offsets to Paint tables, from the start of the LayerV1List table. |
+| uint32 | numLayers |  |
+| Offset32 | paintOffset[numLayers] | Offsets to Paint tables. |
 
-*Note:* For large layer counts PaintComposite can be used to combine multiple COLR v1 glyphs.
+Only layers referenced by `PaintColrLayers` (format 8) records need to 
+be encoded here.
 
 #### Variation structures
 
@@ -563,8 +567,12 @@ If compositeMode value is not recognized, COMPOSITE_CLEAR is used.
 |-|-|-|
 | uint8 | format | Set to 8. |
 | uint8 | numLayers | Number of offsets to Paint to read from layers. |
-| Offset32 | layers | Pointer to a series of Offset32 to Paint. |
+| uint32 | firstLayerIndex | Index into the LayerV1List referenced by the COLR v1 header. |
 
+Each layer is COMPOSITE_SRC_OVER previous.
+
+*Note:* uint8 size saves bytes in most cases. Large layer counts can be
+achieved by way of PaintComposite or a tree of PaintColrLayers.
 
 
 #### Composite modes
@@ -633,11 +641,30 @@ COLR glyphs. This is desirable for reusable parts but introduces the
 possibility of a cyclic graph. Implementations should fail if a cycle
 is detected.
 
+*Note:* Cycle detection can be achieved by keeping a set of addresses of visited paints.
+Before processing a paint check if it's address is in the set, if it is we have a cycle
+and should fail. Once done processing a given paint, including children, take it's address
+out of the set. This allows the same paint to be reached repeatedly as long as no
+cycle is formed. Pseudocode:
+
+```
+# called initially with the base glyph paint and an empty set.
+def paint(paint, active_paints)
+  if paint in active_paints
+    fail, we have a cyle
+  add paint to active_paints
+
+  process paint, potentially calling paint again for referenced paints
+
+  remove paint from active_paints
+```
+
 ##### Bounded Layers Only
 
-The `BaseGlyphV1Record` paint must be bounded.
-Implementations must confirm this invariant.
-Unbounded layers must not render.
+The `BaseGlyphV1Record` paint must define a bounded region. That is,
+is must paint within an area for which a bounding box could be
+defined. Implementations must confirm this invariant.
+A `BaseGlyphV1Record` with an unbounded paint must not render.
 
 The following paints are always bounded:
 
@@ -911,17 +938,15 @@ struct PaintComposite
   Offset24<Paint>     backdrop;
 };
 
-// A reference directly into a series of Paint pointers
-typedef UnsizedArrayOf<Offset32<Paint>> PaintList;
-
 // Each layer is COMPOSITE_SRC_OVER previous
 // NOTE: uint8 size saves bytes in most cases and does not
-// preclude use of large layer counts via PaintComposite.
+// preclude use of large layer counts via PaintComposite or a tree
+// of PaintColrLayers.
 struct PaintColrLayers
 {
   uint8               format; // = 8
   uint8               numLayers;
-  Offset32<PaintList> layers;
+  uint32              firstLayerIndex;  // index into COLRv1::layersV1
 }
 
 struct BaseGlyphV1Record
@@ -931,6 +956,9 @@ struct BaseGlyphV1Record
 };
 
 typedef ArrayOf<BaseGlyphV1Record, uint32> BaseGlyphV1List;
+
+// Only layers accessed via PaintColrLayers (format 8) need be encoded here.
+typedef ArrayOf<Offset32<Paint>, uint32> LayerV1List;
 
 struct COLRv1
 {
@@ -942,6 +970,7 @@ struct COLRv1
   uint16                                            numLayersV0;
   // Version-1 additions
   Offset32<BaseGlyphV1List>                         baseGlyphsV1;
+  Offset32<LayerV1List>                             layersV1;
   Offset32<ItemVariationStore>                      varStore;
 };
 
@@ -992,29 +1021,7 @@ Allocate a bitmap for the glyph according to glyf table entry extents for gid
     7) PaintSolid
           SkCanvas::drawColor with color configured
     8) PaintColrLayers
-        Paint each referenced layer
-
-
-While recursing down through paints, keep a set of addresses of visited paints.
-Before processing a paint check if it's address is in the set, if it is we have a cycle
-and should fail. Once done processing a given paint, including children, take it's address
-out of the set. This allows the same paint to be reached repeatedly as long as no
-cycle is formed. Pseudocode:
-
-# called initially with the base glyph paint and an empty set.
-def paint(paint, active_paints)
-  if paint in active_paints
-    fail, we have a cyle
-  add paint to active_paints
-
-  process paint, potentially calling paint again for referenced paints
-
-  remove paint from active_paints
-
-
-
-
- *1 (in the implementation, potentially collapse each pair as PaintComposite (3) with first gid as backdrop, second as src, or just draw on root SkCanvas directly)
+        Paint each referenced layer by performing a)
 ```
 
 ### FreeType
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -871,7 +871,7 @@ struct PaintColrSlice
   uint8               format; // = 5
   uint16              gid;    // shall be a COLR gid
   uint8               firstLayerIndex;
-  uint8               lastLayer;
+  uint8               lastLayerIndex;
 }
 
 struct PaintTransformed
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -567,7 +567,7 @@ If compositeMode value is not recognized, COMPOSITE_CLEAR is used.
 |-|-|-|
 | uint8 | format | Set to 8. |
 | uint8 | numLayers | Number of offsets to Paint to read from layers. |
-| uint32 | firstLayerIndex | Index into the LayerV1List referenced by the COLR v1 header. |
+| uint32 | firstLayerIndex | Index into the LayerV1List. |
 
 Each layer is composited on top of previous with mode COMPOSITE_SRC_OVER.
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -386,7 +386,7 @@ Offsets are always relative to the start of the containing struct.
 | Type | Name | Description |
 |-|-|-|
 | uint16 | glyphID | Glyph ID of the base glyph. |
-| Offset32 | layerListOffset | Offset to LayerV1List table, from start of BaseGlyphsV1List table. |
+| Offset32 | paintOffset | Offset to Paint, typically a `PaintColrLayers` |
 
 *Note:* The glyph ID is not limited to the numGlyphs value in the &#39;maxp&#39; table.
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -524,7 +524,7 @@ Glyph outline is used as clip mask for the content in the Paint subtable. Glyph 
 | uint8 | format | Set to 5. |
 | uint16 | glyphID | Virtual glyph ID for a BaseGlyphV1List base glyph. |
 | uint8 | firstLayer | First layer to take from the glyph identified by glyphID  |
-| uint8 | lastLayer | Last layer to take from the glyph identified by glyphID |
+| uint8 | lastLayerIndex | Index of the last layer to take from the glyph identified by glyphID |
 
 Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs.
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -25,7 +25,7 @@ December 2019
             - [Radial Gradient](#radial-gradient)
         - [Transformation](#transformation)
         - [Composition](#composition)
-        - [COLR Glyph](#colr-glyph)
+        - [COLR Slice](#colr-slice)
 - [OFF Changes](#off-changes)
     - [Data types](#off-43-data-types)
     - [COLR table](#off-5711-colr--color-table)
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -517,12 +517,14 @@ For linear gradient without skew, set x2,y2 to x1,y1.
...
 Glyph outline is used as clip mask for the content in the Paint subtable. Glyph ID must be less than the numGlyphs value in the &#39;maxp&#39; table.
 
-##### PaintColrGlyph table (format 5)
+##### PaintColrSlice table (format 5)
 
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 5. |
 | uint16 | glyphID | Virtual glyph ID for a BaseGlyphV1List base glyph. |
+| uint8 | firstLayer | First layer to take from the glyph identified by glyphID  |
```
anthrotype, line 526:
```suggestion
| uint8 | firstLayerIndex | Index of the first layer to take from the glyph identified by glyphID  |
```

colr-gradients-spec.md
```diff
@@ -517,12 +517,14 @@ For linear gradient without skew, set x2,y2 to x1,y1.
...
+| uint8 | lastLayer | Last layer to take from the glyph identified by glyphID |
```
anthrotype, line 527:
```suggestion
| uint8 | lastLayerIndex | Index of the last layer to take from the glyph identified by glyphID |
```

colr-gradients-spec.md
```diff
@@ -864,10 +866,12 @@ struct PaintGlyph
                               // must be less than maxp.numGlyphs
 }
 
-struct PaintColrGlyph
+struct PaintColrSlice
 {
   uint8               format; // = 5
   uint16              gid;    // shall be a COLR gid
+  uint8               firstLayer;
```
anthrotype, line 873:
```suggestion
  uint8               firstLayerIndex;
```

colr-gradients-spec.md
```diff
@@ -864,10 +866,12 @@ struct PaintGlyph
...
+  uint8               lastLayer;
```
anthrotype, line 874:
```suggestion
  uint8               lastLayerIndex;
```

colr-gradients-spec.md
anthrotype: Even if it may seem obvious, we should clarify that these indices follow zero-based numbering.

Also, we should say what client should do if the index is invalid (last > first, or if out of bounds). Probably the only thing to do would be to discard the invalid PaintColrSlice.

anthrotype (approved):
LGTM, with comments

colr-gradients-spec.md
anthrotype: > if out of bounds

it's an unsigned 8-bit integer, just like LayerV1List.LayerCount, so it's guaranteed to always be in-bound.

anthrotype:
we should also specify that the `lastLayerIndex` in the range is meant to be inclusive. E.g. a range 0...255 means "include all layers, from first to the last"; or a range `0...0` would mean include only the first layer (unlikely, but just to make the point of last being inclusive).

colr-gradients-spec.md
anthrotype: Let me correct myself again.
Say a base color glyph only has n=5 layers; if the requested layer range max exceeds the available layers (`lastLayerIndex >= n`), I believe the client should _not_ reject the PaintColorSlice, but it should _clamp_ the lastLayerIndex value to `n-1`.

anthrotype:
Should it not be `PaintColorSlice` instead of `PaintColrSlice`?
In the fontTools implementation, I originally called it `PaintColorGlyph`, not `PaintColrGlyph`.
Even in the current draft, we seem to oscillate between `Colr` and `Color`.
I would suggest using `Color` instead of the condensed form `Colr`. Unlike the table tag, there is no reason to make it 4-character logn in this case. It just makes it look like a typo.

anthrotype:
perhaps `PaintColorGlyphSlice` would be clearer, as it's as if we're taking a "slice of a color glyph", not a "slice of color" per se, whatever that means. Ok I'm done with the bikeshed.

rsheeter:
Per f2f, keeping PaintColrSlice to indicate a slice of a COLR glyph.

rsheeter:
Pushed initial pass at updating per further discussion. Tl;dr new format with Offset32 => series of Paint pointers + number of them to use.

colr-gradients-spec.md
```diff
@@ -885,18 +906,25 @@ struct PaintComposite
   Offset24<Paint>     backdrop;
 };
 
+// Each layer is COMPOSITE_SRC_OVER previous
+struct PaintColrLayers
+{
+  uint8               format; // = 8
+  Offset32<LayerList> layers;
+  uint8               numLayers;
```
anthrotype, line 914: The `LayerV1List` is already an `ArrayOf<Offset32<Paint>, uint8>`, where the array of offsets is immediately preceded  a (uint8) count value. What is this `numLayers` for? If it is the lenght of the array, then it would be duplicated.

colr-gradients-spec.md
```diff
@@ -517,14 +525,23 @@ For linear gradient without skew, set x2,y2 to x1,y1.
 
 Glyph outline is used as clip mask for the content in the Paint subtable. Glyph ID must be less than the numGlyphs value in the &#39;maxp&#39; table.
 
-##### PaintColrGlyph table (format 5)
+##### PaintColrSlice table (format 5)
```
anthrotype, line 528: did you mean to undo this change? I thought you wanted to restore the current PaintColrGlyph and add a new paint format 8 for a list of layers?

colr-gradients-spec.md
```diff
@@ -517,14 +525,23 @@ For linear gradient without skew, set x2,y2 to x1,y1.
...
 
-Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs.
+Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs. If firstLayerIndex is greater than the number of layers
+available in the base glyph or lastLayerIndex < firstLayerIndex
+the `PaintColrSlice` is invalid. If lastLayerIndex > the number of
+layers available in the base glyph then all layers from firstLayerIndex
+to the last available layer should be retained.
+
+A range 0..255 therefore means "all available layers." A range 0..0 means
+just the first layer.
```
anthrotype, line 544: I guess the whole paragraph above should be removed now

colr-gradients-spec.md
```diff
@@ -306,13 +307,20 @@ and combined with `backdrop` given the blending rule specified in
 `mode`. Compositing modes are taken from Compositing modes are taken from the
 [W3C Compositing specification](https://www.w3.org/TR/compositing-1/).
 
-## COLR Glyph
+## COLR Glyph 
 
-`PaintColrGlyph` is a special paint which allows reuse of a COLR glyph of this
+`PaintColrGlyph` is a special paint which allows reuse of a COLR glyph of this  
 proposed exension as a paint. Painting a `PaintColorGlyph` means executing the
```
anthrotype, line 313: we need to fix this terminology confusion between `PaintColorGlyph` and `PaintColrGlyph`. Let use the latter only

colr-gradients-spec.md
```diff
@@ -306,13 +307,20 @@ and combined with `backdrop` given the blending rule specified in
...
 paint operations that are described by the `BaseGlyphV1Record` matching the
-glyph id `gid` specified in `PaintColorGlyph`. See section [Reusable
-Parts](#reusable-parts).
+glyph id `gid` specified in `PaintColorGlyph`. See section [Reusable Parts](#reusable-parts).  
+
+## COLR Layers
+
+`PaintColrLayers` is a special paint which allows reuse of some or all of a COLR
+v1 glyph as a paint. It points into a list of Paint pointers and specifies how many
```
anthrotype, line 320: from your definition, it doesn't seem to "point _into_ a list" but rather "point _to_" a list, and the list struct (`ArrayOf`) has its own count value just before the array of offsets.

colr-gradients-spec.md
anthrotype: is `numLayers` uint8 at the end of PaintColrLayers meant to specify how many layers to consume, to be able to reuse partial COLR Glyphs?
That means one can only share the initial numLayers from a LayerV1List -- what about if I want to reuse some other slice that doesn't start at the first layer?

anthrotype: I think we can just get rid of this `numLayers` (perhaps making this an `Offset24<LayerV1List>` to align to 2-byte boundaries).
When you want to reuse a list of layers you create a _new_ LayerV1List and point your base glyph (or some other parent paint) to it. The list will already contain an ArrayOf offsets to Paints, which comes its own count, thus no need for the extra numLayers.

If you really want to be able to take a slice of an existing array of Paints (instead of making a new list), you would need both a start index and a count (or an end index).

colr-gradients-spec.md
rsheeter: Done

colr-gradients-spec.md
rsheeter: Updated, I believe it now does what I actually wanted, ty :)

colr-gradients-spec.md
rsheeter: Fixed now, I think

colr-gradients-spec.md
```diff
@@ -606,16 +628,16 @@ be noted.
...
 ##### Acyclic Graphs Only
 
-`PaintColrGlyph` allows recursive composition of COLR glyphs. This
-is desirable for reusable parts but introduces the possibility of
-a cyclic graph. Implementations should track the COLR gids they have
-seen in processing and fail if a gid is reached repeatedly.
+`PaintColrGlyph` and `PaintColrLayers` allow recursive composition of
+COLR glyphs. This is desirable for reusable parts but introduces the
+possibility of a cyclic graph. Implementations should fail if a cycle
+is detected.
```
drott, line 634: > Implementations should fail if a cycle is detected.

I would suggest to make this stricter and spell it out, similar to as in the pseudocode, for example:

Implementations must not render anything for a COLRv1 if a cycle is detected during processing. A cycle occurs when traversing down a branch of the paint graph, a Paint is referenced a second time after already having been visited.




- during the processing of `PaintColrGlyph` the a `PaintColrGlyph` is encountered that points to the same `glyphID`.
- during the processing of `PaintColrLayers` a `PaintColrLayers` is encountered that points to the

colr-gradients-spec.md
```diff
@@ -960,6 +991,28 @@ Allocate a bitmap for the glyph according to glyf table entry extents for gid
           (expected to be bounded by parent composite mode or clipped by current clip, check bounds?)
     7) PaintSolid
           SkCanvas::drawColor with color configured
+    8) PaintColrLayers
+        Paint each referenced layer
```
drott, line 995: Paint each referenced layer's paint, by performing a).

colr-gradients-spec.md
```diff
@@ -960,6 +991,28 @@ Allocate a bitmap for the glyph according to glyf table entry extents for gid
...
+
+
```
drott, line 997: I suggest to add a heading:

**Cycle detection**

colr-gradients-spec.md
```diff
@@ -606,16 +628,16 @@ be noted.
...
+COLR glyphs. This is desirable for reusable parts but introduces the
+possibility of a cyclic graph. Implementations should fail if a cycle
+is detected.
 
 ##### Bounded Layers Only
 
-Every entry in the `LayerV1List` must define a bounded region.
-Implementations must confirm this invariant. Unbounded layers must
-not render.
+The `BaseGlyphV1Record` paint must be bounded.
```
drott, line 638: Suggest to clarify bounded. Previous phrasing had at least "bounded region", for example:

The `BaseGlyphV1Record` paint must be bounded, which means it must only define a pixel color values for a limited area.

colr-gradients-spec.md
```diff
@@ -309,10 +310,20 @@ and combined with `backdrop` given the blending rule specified in
...
+glyph id `gid` specified in `PaintColrGlyph`. See section [Reusable Parts](#reusable-parts).  
+
+## COLR Layers
+
+`PaintColrLayers` is a special paint which allows reuse of some or all of a COLR
+v1 glyph as a paint. It points to a sequence of Paint pointers and specifies how many
+to consume. This allows reuse of partial COLR Glyphs. For example, suppose glyph
+A has 10 layers, 3 of which are a common backdrop that glyph B also uses. Glyph B
+can define a PaintColrLayers record that points into A's layers and specifies how
+many to take.
```
drott, line 324: I suggest to move the example to a _note_, or _example_  paragraph.

colr-gradients-spec.md
```diff
@@ -960,6 +991,28 @@ Allocate a bitmap for the glyph according to glyf table entry extents for gid
...
+  add paint to active_paints
+
+  process paint, potentially calling paint again for referenced paints
+
+  remove paint from active_paints
+
+
+
 
  *1 (in the implementation, potentially collapse each pair as PaintComposite (3) with first gid as backdrop, second as src, or just draw on root SkCanvas directly)
```
drott, line 1017: This "footnote" is no longer referenced and can be removed.

drott (commented):
Thanks for incorporating what we had discussed, mostly LGTM, but a few comments.

colr-gradients-spec.md
```diff
@@ -545,6 +557,16 @@ Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs.
 
 If compositeMode value is not recognized, COMPOSITE_CLEAR is used.
 
+##### PaintColrLayers table (format 8)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 8. |
+| uint8 | numLayers | Number of offsets to Paint to read from layers. |
+| Offset32 | layers | Pointer to a series of Offset32 to Paint. |
```
anthrotype, line 566: all the offset fields in COLRv0 table currently end with "Offset", so I would suggest to rename this to `paintListOffset`

```suggestion
| Offset32 | paintListOffset | Offset (from beginning of PaintColrLayers table) to PaintList. |
```

colr-gradients-spec.md
```diff
@@ -885,18 +911,23 @@ struct PaintComposite
   Offset24<Paint>     backdrop;
 };
 
-// Layer DAG
+// A reference directly into a series of Paint pointers
+typedef UnsizedArrayOf<Offset32<Paint>> PaintList;
```
anthrotype, line 915: Maybe this should be called `PaintOffsetList`, or `PaintOffsetArray`; it's not really a list of Paints but a list of offsets to Paint.

anthrotype: Also we should clarify what these Offset32 in the PaintList are relative to. In fonttools otTables, offsets are always relative to the parent table; in this case, they would be relative to the PaintList table itself which contains the array. Though it doesn't look like a "table", PaintList is actually a table containing a single element, i.e. the unsized array of offsets to Paint.

rsheeter: Hm, per IM the rule about offset relative to containing struct is going to fall apart when we try to point partway into a sequence of offsets

anthrotype: I think we should have a global PaintList (really a PaintOffsetArray) inlined at the end of COLRv1 header or pointed to by an offset from the top-level table; then each PaintColrLayers would have both a firstPaintIndex and numPaints so it can refer to a slice of that array of offsets to Paint.

colr-gradients-spec.md
rsheeter: Done

colr-gradients-spec.md
rsheeter: Done

colr-gradients-spec.md
rsheeter: I think you can operate purely on Paint and ignore glyphID? - will add a clearer definition. In fact, maybe I'll just move the cycle detection bit up from Rendering to here.

colr-gradients-spec.md
rsheeter: I think this goes away with the global root list

colr-gradients-spec.md
rsheeter: Fixed.

colr-gradients-spec.md
rsheeter: First pass at updating per that comment uploaded, PTAL.

colr-gradients-spec.md
rsheeter: Done

colr-gradients-spec.md
drott: Thanks for the update and moving it here. I think you're right that it's sufficient to operate on `Paints`. How about turning _should fail_ into a _must fail_?

colr-gradients-spec.md
```diff
@@ -606,16 +636,35 @@ be noted.
...
+  remove paint from active_paints
+```
 
 ##### Bounded Layers Only
 
-Every entry in the `LayerV1List` must define a bounded region.
-Implementations must confirm this invariant. Unbounded layers must
-not render.
+The `BaseGlyphV1Record` paint must define a bounded region. That is,
+is must paint within an area for which a bounding box could be
```
drott, line 665: is, is = is, it

Perhaps: "for which a _finite_ bounding box" to make it clear that the bounding box can't have ∞ in its coordinates.

colr-gradients-spec.md
```diff
@@ -545,6 +561,20 @@ Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs.
...
 
+##### PaintColrLayers table (format 8)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 8. |
+| uint8 | numLayers | Number of offsets to Paint to read from layers. |
+| uint32 | firstLayerIndex | Index into the LayerV1List referenced by the COLR v1 header. |
+
+Each layer is COMPOSITE_SRC_OVER previous.
```
drott, line 572: Perhaps reads easier if we say: "Each layer is composited on top of the previous with mode COMPOSITE_SRC_OVER", ditto for in the structs section?

drott (approved):
LGTM with remaining comments.

colr-gradients-spec.md
```diff
@@ -380,10 +394,11 @@ Offsets are always relative to the start of the containing struct.
 
 | Type | Field name | Description |
 |-|-|-|
-| uint8 | numLayers |  |
-| Offset32 | paintOffset[numLayers] | Offsets to Paint tables, from the start of the LayerV1List table. |
```
anthrotype, line 384: I guess you could have kept the part `, from the start of the LayerV1List table.` in here, it's still valid.

colr-gradients-spec.md
```diff
@@ -885,22 +938,28 @@ struct PaintComposite
...
+  uint8               numLayers;
+  uint32              firstLayerIndex;  // index into COLRv1::layersV1
+}
 
-// Each layer is COMPOSITE_SRC_OVER previous
 struct BaseGlyphV1Record
 {
   uint16                gid;
-  Offset32<LayerV1List> layers;
+  Offset32<Paint>       paint;  // Typically PaintColrLayers
```
anthrotype, line 955: you updated the c++ definition for BaseGlyphV1Record but forgot to also update the table layout in the preceding section. BaseGlyphV1Record still is pointing to LayerV1List in the table definition.

colr-gradients-spec.md
```diff
@@ -545,6 +561,20 @@ Glyph ID must be in the BaseGlyphV1List; may be greater than maxp.numGlyphs.
 
 If compositeMode value is not recognized, COMPOSITE_CLEAR is used.
 
+##### PaintColrLayers table (format 8)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 8. |
+| uint8 | numLayers | Number of offsets to Paint to read from layers. |
+| uint32 | firstLayerIndex | Index into the LayerV1List referenced by the COLR v1 header. |
```
anthrotype, line 570: i believe this is redundant, there's only one LayerV1List now

```suggestion
| uint32 | firstLayerIndex | Index into the LayerV1List. |
```

colr-gradients-spec.md
rsheeter: I like it, ty

colr-gradients-spec.md
rsheeter: Done, ty

anthrotype (approved):
LGTM with comments

colr-gradients-spec.md
rsheeter: I prefer not because then it's inconsistently remarked upon and we do have the global rule about always relative to containing struct.

colr-gradients-spec.md
rsheeter: Oh ty, fixing...

PR #89, opened 2020-11-03
misc tweaks

PeterConstable:
Main changes are in description of Affine2x3. As @drott commented in #85,

> as the matrix does not contain basis vectors, but coefficients

So `x-part of x-basis vector` is not really correct. (By convention, the x-part of the x-basis vector would always be 1! The matrix coefficients related to how the basis vector _gets transformed_.)

I separated `_v' = M * v =...` into two lines for better clarity. (The "," gets lost.)

`def paint` isn't prototypical pseudocode; it's Python, and could be unfamiliar to many readers.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -395,7 +395,7 @@ Offsets are always relative to the start of the containing struct.
 | Type | Field name | Description |
 |-|-|-|
 | uint32 | numLayers |  |
-| Offset32 | paintOffset[numLayers] | Offsets to Paint tables. |
+| Offset32 | paintOffsets[numLayers] | Offsets to Paint tables. |
 
 Only layers referenced by `PaintColrLayers` (format 1) records need to be
 encoded here.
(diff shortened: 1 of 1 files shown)


rsheeter:
I think you can find a view in which they are indeed basis vectors of a coordinate system I think the change overall helps, perhaps particularly splitting computation into two parts. +1 to less pythonic pseudocode :)

Nit: I suggest showing computation of a transformed point rather than a vector as this is the typical application in this context. That is,

```
x` = <blah blah>
y` = <blah blah>
```

I think I'm fine with this, @drott, @anthrotype  any concern?

colr-gradients-spec.md
```diff
@@ -620,21 +620,24 @@ Supported composition modes are taken from the W3C [Compositing and Blending Lev
 
 | Type | Name | Description |
 |-|-|-|
-| VarFixed | xx | x-part of x-basis vector |
-| VarFixed | yx | y-part of x-basis vector |
-| VarFixed | xy | x-part of y-basis vector |
-| VarFixed | yy | y-part of y-basis vector |
+| VarFixed | xx | x-component of transformed x-basis vector (*î*) |
```
drott, line 623: Since we talk about î and ĵ only in the note, not in the normative text, I don't think they should be in the description here.

drott (commented):
paintOffsets change, and pseudocode change LGTM, thanks.

I also find the separation of v's x and y component useful.

In general, and in the future, let's land changes for separate concerns in different PRs. If the  î and ĵ issue below is addressed in this PR, I am okay landing all changes in one in this case.

drott:
Thanks for the updated PR and these edits.

Issue #22, opened 2020-06-22
provide example of radial gradient with degenerate transform


anthrotype:
Radial gradients can contain a 2x2 affine matrix to transform the two circles into ellipses, but _without affecting the position of their centres_ -- unlike SVG's `gradientTransform`, or "graphics systems like Skia or Cairo that apply their (often called user-) transformation matrix to everything".
When rendering CORLv1 radial gradients using the latter graphics systems, one should thus apply the _inverse_ matrix to the centres' coordinates to undo the effect of applying the matrix to everything.
But of course some matrices are singular or degenerate (with 0 determinant) and cannot be inverted.
 
The draft spec notes that even these have a "well defined shape", however it is not clear to me what this shape should look like. A visual example could help a lot here.

This is the relevant paragraph:

> Note: Implementations must be careful to properly render radial gradient even if the provided affine matrix is degenerate or near-degenerate. Such radial gradients do have a well-defined shape, which is a strip or a cone filled with a linear gradient. Implementations using the inverse-transform approach noted above can fall back to a linear-gradient combined with a clipping path to achieve proper rendering of problematic affine transforms.

anthrotype:
I'm still trying to figure out what this sentence actually means:

> Implementations using the inverse-transform approach noted above can fall back to a linear-gradient combined with a clipping path to achieve proper rendering of problematic affine transforms

behdad:
> But of course some matrices are singular or degenerate (with 0 determinant) and cannot be inverted.
> 
> The draft spec notes that even these have a "well defined shape", however it is not clear to me what this shape should look like. A visual example could help a lot here.

> I'm still trying to figure out what this sentence actually means:
> 
> > Implementations using the inverse-transform approach noted above can fall back to a linear-gradient combined with a clipping path to achieve proper rendering of problematic affine transforms

This:
![image](https://user-images.githubusercontent.com/142891/86502438-f0eba180-bd57-11ea-98f0-fb0b457f1ff0.png)

Top row is where there's no translate, so we have two circles. The middle row is where there's a transformation matrix that is not degenerate (ie. determinant > 0). The last row is the degenerate case where the ellipse collapses into a line. Now since the two ellipses use the same transformation matrix, it follows that the two lines are parallel, which makes the resulting gradient a linear gradient. The shape of the sides can be enforced using clipping. That's what we mean by "fall back to a linear-gradient combined with a clipping path to achieve proper rendering of problematic affine transforms."

[...]

PR #96, opened 2020-11-09
Clarify edge cases of radial gradients

drott:
Clarify edge cases in note. Remove mention of degenerate radial gradient
transformation matrix as we removed the tranformation matrix from the
gradient definition when we added transforms.

Addresses parts of #22.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -277,15 +277,22 @@ Radial gradients must be rendered by following these steps:
 
 ![Example radial gradient rendering](images/radial_gradients.png)
 
+**Note:** It follows from the algorithm that: a) When both radii are 0, r₀ = r₁
+= 0, then r(ω) is always 0 and nothing is painted. b) If the centers of the
+circles are distinct, one has a zero radius, the other has a non-zero radius,
+the resulting shape resembles a triangle that is open to one side, see Figure 3
+below. c) If the circles centers are identical, one has a zero radius, one has a
+non-zero radius, then the gradient is well-defined and is painted according to
+the algorithm.
+
 *__Figure 3:__ Example of a radial gradient rendering with extend modes pad,
 repeat and reflect (top to bottom) with color stops for blue at 0, yellow at 0.5
 and red at 1. (Illustration generated from <a
 href="images/radial_gradients.svg">images/radial_gradients.svg</a>)*
 
-**Note:** Implementations must be careful to properly render radial gradient
-even if they are subject to a *[degenerate](https://en.wikipedia.org/wiki/Invertible_matrix)*
-or *near-degenerate* transform. Such radial gradients do have a well-defined shape, which is
-a strip or a cone filled with a linear gradient.
+**Note:** When a radial gradient is nested below a transformation which flattens
+the circles so that they resemble lines, a radial gradient may appear as a a
+strip or a cone filled with a linear gradient.
 
 ## Transformation
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -279,11 +279,11 @@ Radial gradients must be rendered by following these steps:
 
 **Note:** It follows from the algorithm that: a) When both radii are 0, r₀ = r₁
 = 0, then r(ω) is always 0 and nothing is painted. b) If the centers of the
-circles are distinct, one has a zero radius, the other has a non-zero radius,
-the resulting shape resembles a cone that is open to one side, see Figure 3
-below. c) If the circles centers are identical, one has a zero radius, one has a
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -277,15 +277,22 @@ Radial gradients must be rendered by following these steps:
 
 ![Example radial gradient rendering](images/radial_gradients.png)
 
+**Note:** It follows from the algorithm that: a) When both radii are 0, r₀ = r₁
+= 0, then r(ω) is always 0 and nothing is painted. b) If the centers of the
+circles are distinct, one has a zero radius, the other has a non-zero radius,
+the resulting shape resembles a triangle that is open to one side, see Figure 3
```
anthrotype, line 283: > resembles a triangle that is open to one side

it looks like a "cone" to me

colr-gradients-spec.md
```diff
@@ -277,15 +277,22 @@ Radial gradients must be rendered by following these steps:
...
 and red at 1. (Illustration generated from <a
 href="images/radial_gradients.svg">images/radial_gradients.svg</a>)*
 
-**Note:** Implementations must be careful to properly render radial gradient
-even if they are subject to a *[degenerate](https://en.wikipedia.org/wiki/Invertible_matrix)*
-or *near-degenerate* transform. Such radial gradients do have a well-defined shape, which is
-a strip or a cone filled with a linear gradient.
+**Note:** When a radial gradient is nested below a transformation which flattens
+the circles so that they resemble lines, a radial gradient may appear as a a
+strip or a cone filled with a linear gradient.
```
anthrotype, line 295: > a cone filled with a linear gradient

this specific case would be worth a visual example. It's not immediately clear to me how a transformation that flattens circles to lines would be equivalent to "a cone filled with a linear gradient".

colr-gradients-spec.md
drott: Changed to cone.

colr-gradients-spec.md
drott: I suggest to add file a new issue for adding an example under transforms. Both circles are flattened so they become lines, if one is bigger than the other, we have a linear gradient between those two lines, which extends outwards on each side to form a cone.

anthrotype: but IIUC, since the same transform is applied to both circles (to both their centres and radii, no longer just to their radii as before), if one is flattened to a line, the other cirlce will also be flatteded _along the same line_ as the other circle. So the gradient disappears to a line, never a cone or trapezoid.

drott: If you imagine a gradient defined by a tall circle and a shorter one, both very far apart, then a transform that horizontally shrinks the whole thing, then the circles become disks, and the gradient becomes denser, but it can still be a cone shape. Example here: https://gist.github.com/drott/49f9a95f2f5b31dd275c191eb12d39db

```
  <defs>
    <radialGradient id="radgrad-pad"
                    gradientUnits="userSpaceOnUse"
                    fx="-5000" fy="0" fr="120"
                    cx="10000" cy="0" r="130"
                    spreadMethod="pad">
      <stop offset="0%" stop-color="#DA0025"/>
      <stop offset="50%" stop-color="#FFC600"/>
      <stop offset="100%" stop-color="#0052AB"/>
    </radialGradient>
    <rect id="filled-rect" x="-300000" y="-200000" width="800000" height="400000" stroke-width="2" stroke="gray"/>
  </defs>
  <g transform="scale(0.001 1)" class="spreadmethod-repeat">
    <use href="#filled-rect" fill="url(#radgrad-pad)"/>
  </g>
```

anthrotype: ok makes sense now, thanks

anthrotype (approved):

rsheeter (approved):
Example strip would be nice, could be separate PR

colr-gradients-spec.md
```diff
@@ -277,15 +277,22 @@ Radial gradients must be rendered by following these steps:
 
 ![Example radial gradient rendering](images/radial_gradients.png)
 
+**Note:** It follows from the algorithm that: a) When both radii are 0, r₀ = r₁
+= 0, then r(ω) is always 0 and nothing is painted. b) If the centers of the
+circles are distinct, one has a zero radius, the other has a non-zero radius,
+the resulting shape resembles a cone that is open to one side, see Figure 3
```
PeterConstable, line 283: There is a further condition to produce the appearance of a cone: the center of the one circle must lie outside the circumference of the other circle.

drott: Thanks, incorporated that into the note.

PR #97, opened 2020-11-09
Insert PaintColrLayers as format number 1

### LINKED ISSUE #90: Make PaintColrLayers format 0?
Given the latest design change, PaintColrLayers basically replaces the LayersV1List tables of the previous design. (The new LayersV1List is a global array; the per-base-glyph data in the previous LayersV1List tables now is held in PaintColrLayers.) That makes PaintColrLayers pretty fundamental in the structure of a color glyph description. From that perspective, it could seem odd, then, that it's the *8th* paint format, not the 1st format. Would it make sense to renumber the formats to make this format 1?

So far, we've avoided having a format 0. Is there a particular reason? If not, a less disruptive change would be to make PaintColrLayers format 0.

drott:
Shift remaining format numbers up. Sort pseudocode loop by formats.

Fixes #90.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -397,8 +397,8 @@ Offsets are always relative to the start of the containing struct.
 | uint32 | numLayers |  |
 | Offset32 | paintOffset[numLayers] | Offsets to Paint tables. |
 
-Only layers referenced by `PaintColrLayers` (format 8) records need to 
-be encoded here.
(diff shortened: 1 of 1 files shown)


rsheeter:
LGTM

anthrotype (approved):
LGTM

rsheeter (approved):

PR #98, opened 2020-11-09
Add example of first circle contained in second

### LINKED ISSUE #22: provide example of radial gradient with degenerate transform

drott:
Fixes #22.


drott:
Preview link: https://github.com/drott/colr-gradients-spec/tree/radialContainedExample

rsheeter (approved):

PeterConstable:
I had already generated svg/png for this, but hadn't yet gotten to creating a PR to add to this repo. It was in the PDF I sent you a few days ago.

drott:
> I had already generated svg/png for this, but hadn't yet gotten to creating a PR to add to this repo. It was in the PDF I sent you a few days ago.

Thanks - it was quickly done so I thought I make those so we can make progress and close some issues. As this PR is ready to merge, do you have a preference which ones to land? I am fine either way as long as we have a separate PR for them.

PeterConstable:
Yours and mind have separate advantages:
- yours has the circles closer to the horizontal center, so more of the extend on the right side is visible.
- in mind, I moved the label for c0 so that less of the extend gradation within the small circle was covered.
I'll rev mind to incorporate the advantage of yours.

drott:
Merged #111 instead.

PR #103, opened 2020-11-11
For OFF amendments, chg W3C Comp & Blend to norm've ref

PeterConstable:


drott (approved):
LGTM

@rsheeter @anthrotype merge?

PeterConstable:
@drott @rsheeter @anthrotype Can this get merged?

rsheeter:
I think I had this normative and was advised to make it not so, matching how other W3 content is referenced.

rsheeter:
I made it non-normative per discussion with Vlad so Is suggest we leave the compositing non-normative. Adding the other ref makes sense to me.

PeterConstable:
Vlad has said "Candidate Recommendations" can't be used as normative references, but the ISO Directives don't say that.  The content of the W3C spec is absolutely required for implementations, so for OFF to be complete, if it's not in a normative reference, then the relevant bits would need to be copied out of the W3C spec into OFF. 

I think the best approach is to list this as a normative reference; Vlad can investigate with the ISO Central Secretariat if there's a problem and if so we can work out how best to resolve at that point.

rsheeter:
@vlevantovsky, I believe you advised me not to make this normative. Please advise.

vlevantovsky:
Correct. I would need to double-check with the ISO Editor to be sure, but in the past they rejected normative references to Candidate Recommendation, only final published versions were allowed to be listed as normative. (This is the reason why SVG1.1 is a normative reference in OFF, and SVG2 is informative.)
If they tell me it's ok to do now, we can easily change it in the future.

PeterCon:
In the mean time, I think the doc should handle the reference the way we think it should be proposed to SC29, whether or not the editor is able to incorporate it that way.

vlevantovsky:
We can also treat it as an external constraint, and designate the implementation of the composition modes as a "must". The text of the draft you initially proposed had this language (with no designation of the referenced spec as either normative or informative):

> Supported composition modes are taken from the W3C Compositing and Blending Level 1 specification. In Paint tables, a composition mode is specified using the following enumeration.

I don't see how our current language is different conceptually, and I think we'd be in a better position if we make a strong emphasis on the importance of this reference in the spec text, rather than relying on a designation of the reference as either normative or informative.

rsheeter:
> in the past they rejected normative references to Candidate Recommendation

That's enough to make it informative for me, particularly given that I don't perceive there being any real disadvantage to it.

PeterCon:
>I don't perceive there being any real disadvantage to it.

In ISO-ese, it's quite different.

That said, the *Normative references* clause is, itself *informative*, "given for the convenience of the user".

It should still be listed as a normative reference if ISO will accept it.

PeterCon:
>We can also treat it as an external constraint, and designate the implementation of the composition modes as a "must". 

That's not really consistent with the ISO Directives. 

>External constraints are not requirements of the document. They are given for the information of the user.

>Use of the word “must” does not imply that the external constraint referred to is a requirement
(3.3.3) of the document.

vlevantovsky (changes requested):
According to the information received from the ISO Editor, normative references to W3C Candidate Recommendation documents are not allowed. Please keep it as informative reference, however we can add some language about this reference to make it clear that this is essential for COLR v1 implementations.

PeterConstable:
Per Vlad's feedback, the entry for the Composition & Blending doc is proposed for addition to the bibliography, not the Normative References clause.

colr-gradients-spec.md
```diff
@@ -382,6 +382,8 @@ We're proposing changes to the following sections of ISO/IEC 14496-22:2019 “OF
 
 An overview of the design is provided, followed by the suggested specific changes.
 
+_Add the following reference:_
+
```
rsheeter, line 386: dangling, remove?

rsheeter (approved):

PR #105, opened 2020-11-11
Add note to PaintColrGlyph contrasting with PaintColrLayers

PeterConstable:


drott:
Thanks, LGTM. I took the freedom to line-break the note and add markdown formatting, hope that's okay.

drott (approved):

PeterCon:
NP. I had "NOTE" formatted the way it would appear in OFF, but that's a minor detail.

drott:
Ok, I realized we have less consistent formatting of the Notes paragraphs than I thought, so we should probably do a pass to harmonize them.

PeterCon:
Your convention is fine.

PR #107, opened 2020-11-11
Fix Note styling, caption and note order

drott:
Follow-up to #96.


drott:
Merging as minor fix to previous merge.

Issue #88, opened 2020-10-31
Rotated components


Lorp:
In the context of COLRv1 being a superset of the features of composite glyphs, I wonder if it is too late to consider adding the possibility of rotation by a numeric value, rather than using the affine transformation matrix? Especially when combined with a variation axis, it would be useful not just for arrows, ornaments and clock faces, but also to enable chained rotations, so that mechanical objects, such as animal skeletons, can be represented efficiently. Such freedom of movement must currently be represented by multiple keyframes or by higher order interpolation (HOI). The former suffers from inaccuracy, the latter from an opaque build process and lack of support in UI.

arrowtype:
YES. Proper rotation would be huge!

PeterCon:
Better support for rotation in variation (continuous variation of an angle θ) sounds useful, but it doesn't seem like burying it within the COLR table would be exactly what you want. I mean, it would be possible to have behaviour like a non-colour font—the glyph fill is only ever text foreground colour; but needing to add COLR and CPAL to get that seems kludgy. 

What makes rotation as variation difficult now is that the basis function used to calculate variation scalar coefficients is strictly linear (in a single active segment—a "hockey stick" function—or in two segments going from 0 to 1 to 0—a "tooth" function). What if the format were extended to support non-linear basis functions?

Lorp:
Future-proofing this new format, with capabilities that are called for in ‘glyf’, is reasonable I think. COLRv1 in fact already does this to an extent by enabling proper handling of white-on-black shapes, which is nothing to do with “colour”. As @davelab6 wrote here https://github.com/googlefonts/colr-gradients-spec/issues/3#issuecomment-694888194 : “I don't see an issue with CPAL v2 [_sic._] edging ahead, and then updating glyf to match the capability set”.

So I would propose rotation in turns or degrees. I would submit that rotation and reflection are the most common use of the affine matrix, and should have been in ‘glyf’ all along. Reflection is satisfactorily represented in Affine2x2, but for rotation the representation is poor. If rotation was recorded directly in turns or degrees, many of the most common rotations are represented exactly, without semantic loss and therefore with variation potential.

PeterConstable:
So, perhaps you have in mind something like the following?

*PaintTransformed table:*
| Type | Field name | Description |
|-|-|-|
| uint8 | format | Set to 6. |
| uint8 | flags | See below for flag details. |
[...]

Issue #56, opened 2020-09-29
should Affine2x3.dx/dy be integers instead of Fixed 16.16?


anthrotype:
@Lorp noted on https://github.com/MPEGGroup/OpenFontFormat/issues/20#issuecomment-700765671

> The table at line 170 defines an Affine2x3 matrix with two translation elements, dx and dy. It it intentional that these are VarFixed (16.16) rather than VarFWord (16-bit integer)? It seems to imply that a renderer needs to keep track of fractional point locations.

In glyf table the component x and y offsets are integers, just like the simple glyph point coordinates are. Also in this draft spec, the gradients' end points are encoded as integers (VarFWORD).
Allowing fractional point locations as result of PaintTransform would be inconsistent with that.

PeterConstable:
The final results _at some point in derivation_ needs to be integral design units. The spec should probably be clear as to when: is it as each paint is composed with its parent, or only after the entire graph has been composed?

rsheeter:
> It seems to imply that a renderer needs to keep track of fractional point locations.

Agreed. This is already true when rasterizing and when computing variations. It is desirable when computing a chain of transforms.

> The final results at some point in derivation needs to be integral design units

This is downstream of design space; we've moved on to the rasterizers coordinate system and we aren't coming back to design space.

PeterConstable:
Except that the rasterizers are working with a particular PPEM, whereas the font data is defined in design units.

Lorp:
@rsheeter Are you saying that composite glyphs (`glyf` table) are composed in the rasterizer coordinate system? Thus you could in effect get a 2x ppem by using a 0.5 scale transform.

Lorp:
> we've moved on to the rasterizers coordinate system

Technically there may not be a rasterizer. It seems to me to be an innovation that composite decomposition introduces fractions.

Lorp:
@rsheeter the chaining argument is persuasive. If 100% compatibility between a COLR-style glyph stack and a glyf-based composite glyph is desired (including the potentially common case of representing the b/w fallback glyph using identical base glyphs), the 16-bit glyf matrix values can be used, padded with zeroes from F2.14 or DWord to 32-bit Fixed.

behdad:
Non-integer coordinates are already in the format, via transformed composite glyphs in `glyf` table, non-integer numbers in `CFF`/`CFF2`, as well as arbitrary decimal numbers in `SVG`.

PR #113, opened 2020-11-13
add alternate transform representation in PaintTransformed

PeterConstable:
Per #88

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -630,7 +630,7 @@ benefits.*
 |-|-|-|
 | uint8 | format | Set to 8. |
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
-| VarFixed | angle | Rotation angle, in clockwise degrees. |
+| VarFixed | angle | Rotation angle, in counter-clockwise degrees. |
 | VarFWord | centerX | x coordinate for the center of rotation. |
 | VarFWord | centerY | y coordinate for the center of rotation. |
 
@@ -642,8 +642,8 @@ benefits.*
 |-|-|-|
 | uint8 | format | Set to 9. |
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
-| VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in clockwise degrees. |
-| VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in clockwise degrees. |
+| VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in counter-clockwise degrees. |
+| VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in counter-clockwise degrees. |
 
 *__Note:__ Skews can also be represented using the PaintTransformed table. The important difference is in being able to specify skew as an angle rather than as changes to basis vectors.*
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -1165,9 +1165,26 @@ struct PaintTransformed
   Affine2x3           transform;
 };
 
-struct PaintComposite
+struct PaintRotate
 {
   uint8               format; // = 8
+  Offset24<Paint>     src;
+  VarFixed            angle
+  VarFWord            centerX
+  VarFWord            centerY
+};
+
+struct PaintSkew
+{
+  uint8               format; // = 9
+  Offset24<Paint>     src;
+  VarFixed            xSkewAngle
+  VarFixed            ySkewAngle
+};
+
+struct PaintComposite
+{
+  uint8               format; // = 10
   Offset24<Paint>     src;
   CompositeMode       mode;   // If mode is unrecognized use COMPOSITE_CLEAR
   Offset24<Paint>     backdrop;
@@ -1242,7 +1259,17 @@ Allocate a bitmap for the glyph according to glyf table entry extents for gid
           apply transform
           call a) for paint
           restore
-    8) PaintComposite
+    8) PaintRotate
+          saveLayer()
+          apply transform
+          call a) for paint
+          restore
+    9) PaintSkew
+          saveLayer()
+          apply transform
+          call a) for paint
+          restore
+    10) PaintComposite
           paint Paint for backdrop, call a)
           saveLayer() with setting composite mode, on SkPaint
           paint Paint for src, call a)
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -629,7 +629,7 @@ benefits.*
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 8. |
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
 | VarFixed | angle | Rotation angle, in counter-clockwise degrees. |
 | VarFixed | centerX | x coordinate for the center of rotation. |
 | VarFixed | centerY | y coordinate for the center of rotation. |
@@ -641,11 +641,11 @@ benefits.*
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 9. |
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintSkew table. |
 | VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in counter-clockwise degrees. |
 | VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in counter-clockwise degrees. |
-| VarFixed | centerX | x coordinate for the center of rotation. |
-| VarFixed | centerY | y coordinate for the center of rotation. |
+| VarFixed | centerX | x coordinate for the center of skew. |
+| VarFixed | centerY | y coordinate for the center of skew. |
 
 *__Note:__ Skews can also be represented using the PaintTransformed table. The important difference is in being able to specify skew as an angle rather than as changes to basis vectors. Also, when varying angles, a representation directly in degrees is more amenable to smooth variation.*
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -631,8 +631,8 @@ benefits.*
 | uint8 | format | Set to 8. |
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
 | VarFixed | angle | Rotation angle, in counter-clockwise degrees. |
-| VarFWord | centerX | x coordinate for the center of rotation. |
-| VarFWord | centerY | y coordinate for the center of rotation. |
+| VarFixed | centerX | x coordinate for the center of rotation. |
+| VarFixed | centerY | y coordinate for the center of rotation. |
 
 *__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing angle to be specified directly in degrees, which is more amenable to smooth variation.*
 
@@ -1170,8 +1170,8 @@ struct PaintRotate
   uint8               format; // = 8
   Offset24<Paint>     src;
   VarFixed            angle
-  VarFWord            centerX
-  VarFWord            centerY
+  VarFixed            centerX
+  VarFixed            centerY
 };
 
 struct PaintSkew
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -644,6 +644,8 @@ benefits.*
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
 | VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in counter-clockwise degrees. |
 | VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in counter-clockwise degrees. |
+| VarFixed | centerX | x coordinate for the center of rotation. |
+| VarFixed | centerY | y coordinate for the center of rotation. |
 
 *__Note:__ Skews can also be represented using the PaintTransformed table. The important difference is in being able to specify skew as an angle rather than as changes to basis vectors.*
 
@@ -1180,6 +1182,8 @@ struct PaintSkew
   Offset24<Paint>     src;
   VarFixed            xSkewAngle
   VarFixed            ySkewAngle
+  VarFixed            centerX
+  VarFixed            centerY
 };
 
 struct PaintComposite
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -608,8 +608,22 @@ benefits.*
...
+| uint8 | flags | See below for flag details. |
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintTransformed table. |
-| Affine2x3 | transform | An Affine2x3 record (inline). |
+| Affine2x3 or<br>VarFixed | transform | An Affine2x3 record (inline). |
+
+The following flags are defined:
+
+| Mask | Name | Description |
+|-|-|-|
+| 0x01 | TRANSFORM_AFFINE_2X3 | The transform field is an Affixe2x3 record. |
```
anthrotype, line 619: can't we use a value of zero `0b000` to mean the default "TRANSFORM_AFFINE_2X3". We'd save one bit, and automatically make mutually exclusive with the case when another bit is set.

colr-gradients-spec.md
```diff
@@ -608,8 +608,22 @@ benefits.*
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 7. |
+| uint8 | flags | See below for flag details. |
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintTransformed table. |
```
anthrotype, line 612: if we add a uint8 for the flags after the uint8 format, then it is no longer necessary to have a 3-byte Offset24, but we can have a full Offset32 if we still care about the even alignment.

colr-gradients-spec.md
```diff
@@ -608,8 +608,22 @@ benefits.*
...
-| Affine2x3 | transform | An Affine2x3 record (inline). |
+| Affine2x3 or<br>VarFixed | transform | An Affine2x3 record (inline). |
```
anthrotype, line 613: does this "or" make the PaintTransform table lenght depend on the content of "flags"?
I don't like this, I prefer that each paint format has a constant and predefined length. We can define as many paint formats as we need (within 256 limit), so how about we define additional PaintTransform variants for the rotation and skew degrees?

PeterConstable: Yes, as noted in the discussion of #88. There is a precedent, the value record used in GPOS. But it is a downside of this proposal.

If a different format, we should decide on one VarFixed field with a flag to indicate rotate, skew x or skew y; or three VarFixed fields.

rsheeter:
As noted on #88 IMHO we should do a new Paint with one field: angle. Skew and scale and so on are best represented by PaintTransformed; you can Rotate>PaintTransformed (and vice versa) if you need both.

colr-gradients-spec.md
PeterConstable: Per discussion in #88, reverting PaintTransform and adding separate rotate/skew formats.

colr-gradients-spec.md
PeterConstable: Per discussion in #88, reverting PaintTransform and adding separate rotate/skew formats.

colr-gradients-spec.md
PeterConstable: Per discussion in #88, reverting PaintTransform and adding separate rotate/skew formats. So, no more flags.

colr-gradients-spec.md
```diff
@@ -611,11 +611,34 @@ benefits.*
...
+#### PaintAngularSkew table (format 9)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 9. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in clockwise degrees. |
+| VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in clockwise degrees. |
+
+*__Note:__ Skews can also be represented using the PaintTransformed table. The important difference is in being able to specify skew as an angle rather than as changes to basis vectors.*
```
rsheeter, line 635: Nit: suggest similar language to PaintRotate about the key being smooth variable of angle. Could even mention "such as for slnt"

colr-gradients-spec.md
```diff
@@ -611,11 +611,34 @@ benefits.*
...
+
+##### PaintComposite table (format 10)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 10. |
 | Offset24 | sourcePaintOffset | Offset to a source Paint table, from start of PaintComposite table. |
 | uint8 | compositeMode | A CompositeMode enumeration value. |
 | Offset24 | backdropPaintOffset | Offset to a backdrop Paint table, from start of PaintComposite table. |
```
rsheeter, line 644: Could I impose on you to update the C++ structs section and the pseudocode? - if you prefer not to let me know and I'll do it after we merge this

colr-gradients-spec.md
```diff
@@ -611,11 +611,34 @@ benefits.*
...
 |-|-|-|
 | uint8 | format | Set to 8. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| VarFixed | angle | Rotation angle, in clockwise degrees. |
+| VarFWord | centerX | x coordinate for the center of rotation. |
+| VarFWord | centerY | y coordinate for the center of rotation. |
+
+*__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing angle to be specified directly in degrees, which is more amenable to smooth variation.*
+
+#### PaintAngularSkew table (format 9)
```
anthrotype, line 626: just `PaintSkew` would be clear enough to me. SVG has `skewX(angle)` and `skewY(angle)` transforms.

colr-gradients-spec.md
PeterConstable: Thing is, I don't see how it's more amenable to smooth variation. The problem with rotations and variation is you can't easily vary circular functions sin/cos using linear interpolation. But for skew, it's a matter of varying rotation angle versus varying coordinates of basis vectors, and either is easily done smoothly.

colr-gradients-spec.md
PeterConstable: I can, but was waiting until the churn on design is settled.

colr-gradients-spec.md
```diff
@@ -611,11 +611,34 @@ benefits.*
...
 
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 8. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| VarFixed | angle | Rotation angle, in clockwise degrees. |
+| VarFWord | centerX | x coordinate for the center of rotation. |
+| VarFWord | centerY | y coordinate for the center of rotation. |
+
+*__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing angle to be specified directly in degrees, which is more amenable to smooth variation.*
```
rsheeter, line 624: Thinking "aloud" I wonder if we should show the equivalent matrices for rotate and skew, akin to https://www.w3.org/TR/SVG11/coords.html#RotationDefined.

colr-gradients-spec.md
rsheeter: Given the note explaining why it exists +1.

PeterConstable: That was my first thought. When Rod mentioned "angularskew", it seemed like it provided better differentiation from skew expressed using the matrix. But no strong opinion. Shorter is good.

rsheeter (approved):

rsheeter:
I'm very sleepy but this lgtm. @anthrotype wdyt?

colr-gradients-spec.md
PeterConstable: There are lots of good sources for affine transform matrices. The only thing not clear about rotations expressed as matrices is which direction angles (as argument of sin(), cos()) are measured. The SVG spec just says "angle a", which leaves that unclear. I'm not sure what the benefit would be.

colr-gradients-spec.md
```diff
@@ -611,11 +611,34 @@ benefits.*
...
+
+#### PaintAngularSkew table (format 9)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 9. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
+| VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in clockwise degrees. |
+| VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in clockwise degrees. |
```
anthrotype, line 633: should PaintSkew also have centerX and centerY like PaintRotate?
maybe not, given the most immediate use case is to slant by some degrees around the origin (0,0) -- as in slnt axis.
And it can always be achieved via a translate-only PaintTransform wrapping the PaintSkew

colr-gradients-spec.md
PeterConstable: Changed.

colr-gradients-spec.md
PeterConstable: I'm not sure what "center of rotation" for skew would mean.

rsheeter: Hm, I think for the slant case perhaps you do want that. A 2x3 can skew and then translate to the desired position in one. Here's a crude example, note the translate was needed to permit an arbitrary origin. Without it the whole shape "moves" if you will.

```
<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 128 128" viewBox="0 0 128 128">
  <rect x="25" y="25" width="25" height="80" opacity="0.25" />
  <rect x="25" y="25" width="25" height="80" opacity="0.25" transform="skewX(-15) translate(28,0)"/>
</svg>
```

image

rsheeter: ...why is the image enormous github?!

anthrotype: like the centre of rotation, the centre of skew (implicitly the origin (0,0)) is thing around which you slant or skew something. It's the point that is unaffected by the transformation.
equivalent to  ```transform="translate(-256,-256) skewX(20) translate(256, 256)"``` for a 20 degree skew around point (256, 256)

rsheeter: Maybe encoding it as origin (for the purpose of the skew) would be more intuitive?

colr-gradients-spec.md
rsheeter: It's potentially helpful to an implementation that would probably compute the Affine (after varying the #s) to actually apply the rotation. No real objection to leaving it out tho.

colr-gradients-spec.md
rsheeter: I see it as equivalent; imagine you want to vary between a skew of say 0 and -10. That's a nuisance to express by varying PaintTransformed in just the same way varying an overall rotation is.

colr-gradients-spec.md
Lorp: `centerX` and `centerY` should be `VarFixed`, not `VarFWord` (just as it was [persuasively argued](https://github.com/googlefonts/colr-gradients-spec/issues/56) that `Affine2x3.dx` and `Affine2x3.dy` should be VarFixed.

colr-gradients-spec.md
PeterConstable: I get what it would mean now. But I don't see any examples of frameworks taking center x/y for skew (not in CSS, SVG or Canvas 2D).

As you say, one could wrap the skew in translations (add'l 108 bytes). The same issue applies to the matrix, which skews about the origin. I wonder just how common it would be used if add'l members were added (2 x VarFWord = +12 bytes for every use of PaintSkew).

colr-gradients-spec.md
PeterConstable: In that example, expressing the amount as angles is the benefit, not being able to smooth variation. So, I've mentioned being able to specify as angle, but not that this is more amenable to smooth variation.

colr-gradients-spec.md
rsheeter: Imagine using it for slnt. How often is skew relative to the actual origin what you want? - I think you'll end up wrapping transforms (one before, one after) around the skew. Better to pay the 12 bytes.

The 2x3 matrix can be computed to roll all this together (a matrix equivalent to translate, skew, translate).

colr-gradients-spec.md
PeterConstable: OK, I see the point in your example of varying slnt angle: in the matrix, you're trying to vary cos(angle). Will update.

colr-gradients-spec.md
```diff
@@ -624,11 +624,36 @@ benefits.*
 | Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintTransformed table. |
 | Affine2x3 | transform | An Affine2x3 record (inline). |
 
-##### PaintComposite table (format 8)
+#### PaintRotate table (format 8)
 
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 8. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
```
drott, line 632:
```suggestion
| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
```

colr-gradients-spec.md
```diff
@@ -624,11 +624,36 @@ benefits.*
...
+| VarFixed | centerY | y coordinate for the center of rotation. |
+
+*__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing an angle to be specified directly in degrees, which is more amenable to smooth variation.*
+
+#### PaintSkew table (format 9)
+
+| Type | Field name | Description |
+|-|-|-|
+| uint8 | format | Set to 9. |
+| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotateSkew table. |
```
drott, line 644:
```suggestion
| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintSkew table. |
```

colr-gradients-spec.md
```diff
@@ -624,11 +624,36 @@ benefits.*
...
+| VarFixed | xSkewAngle | Angle of skew in the direction of the x-axis, in counter-clockwise degrees. |
+| VarFixed | ySkewAngle | Angle of skew in the direction of the y-axis, in counter-clockwise degrees. |
+| VarFixed | centerX | x coordinate for the center of rotation. |
```
drott, line 647:
```suggestion
| VarFixed | centerX | x coordinate for the center of skew. |
```

colr-gradients-spec.md
```diff
@@ -624,11 +624,36 @@ benefits.*
...
+| VarFixed | centerY | y coordinate for the center of rotation. |
```
drott, line 648:
```suggestion
| VarFixed | centerY | y coordinate for the center of skew. |
```

drott (commented):
Generally looks good to me, a few suggested edits to update to latest naming.

drott (approved):

drott:
@googlebot I fixed it.

PR #116, opened 2020-11-24
Use "shall" instead of "must" in several places

### LINKED ISSUE #108: "must" be sorted vs. "shall"/"should"
Sorry, I should have commented on the PR before it was completed.

>Entries must be sorted...

In ISO-speak (ISO Directives Part 2), "must" is used for external constraints, i.e., something outside the standard that nonetheless imposes a constraint. Like laws of nature, or legal requirements. By having "must" in the proposed draft, it will leave an open question: _Is there something external that implicitly forces this requirement? Or should this be changed to "shall", or to "should"?_

I don't think there's any external constraint (in the ISO sense). We should decide now if we mean "shall" or "should".

"Shall" would imply that, unless the entries are sorted, the font is not conformant, i.e., invalid. Do we expect implementations to reject a font? (I don't think so.) Or do we expect validation tools to call out an error (vs. warning) and deem the font invalid?

(Of course, we could state as a requirement, but runtime implementations may still choose to be lenient.)

There are many parts of the OT spec that say "[items of some kind] are sorted" (implying expectation) or "... must be sorted" ("must" is OT's "shall"). So there's precedent for stating as a requirement rather than recommendation.

Also, the existing text in 5.7.11 (taken from OT without being translated into ISO-speak) does say wrt base glyph records,

>The base glyph records are sorted by glyph id. It is assumed that a binary search can be used to efficiently
access the glyph IDs that have color values.

So, I suggest either that we adopt OT-speak and say "Entries are sorted...", or adopt ISO-speak and say "Entries shall be sorted..."

rsheeter:
As suggested on #108, fixes #108

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -426,7 +426,7 @@ Offsets are always relative to the start of the containing struct.
 | uint32 | numBaseGlyphV1Records |  |
 | BaseGlyphV1Record | baseGlyphV1Records[numBaseGlyphV1Records] | |
 
-Entries must be sorted in ascending order of the `glyphID` field of the `BaseGlyphV1Record`s.
+Entries shall be sorted in ascending order of the `glyphID` field of the `BaseGlyphV1Record`s.
 
 *__Note:__ The sorted order allows implementations to perform binary search to
 find a matching `BaseGlyphV1Record` for a specific `glyphID`.*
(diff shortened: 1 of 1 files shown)


rsheeter:
@PeterConstable for some reason you aren't in the reviewer list; ptal and comment if this looks good to you or not.

drott (approved):

anthrotype:
> PeterConstable for some reason you aren't in the reviewer list

looks like Peter was invited as collaborator already but the invite shows as "pending"

image

PeterCon:
> > PeterConstable for some reason you aren't in the reviewer list
> 
> looks like Peter was invited as collaborator already but the invite shows as "pending"

I didn't see that invite, and can't find it now. Could you resend?

anthrotype:
try visit https://github.com/googlefonts/colr-gradients-spec/invitations

PeterConstable:
Invitation expired.

rsheeter:
Resent

PR #119, opened 2020-11-28
Restructure doc for addition of content for OFF

PeterCon:
I want to add more content to incorporate into amendments for OFF, including some that currently in the Graphical Primitives / Paints section as well as other new content. I think the cleanest way to do this is to start a new section for the OFF amendment content, and build it up. I've placed this for now at the end of the doc, to keep things clear.

I've created commits for some distinct steps, so that people can look at commit-by-commit diffs to see the changes. 

* 53d0598 just sets up the new section, and replaces the intro in the existing OFF section with a new intro. The only significant TOC change is the link for the new 
* 671307f replaces the content for OFF 4.3 with the complete statement required for the amendment (matches was was in the WD presented to WG3).
* 9cf6513 moves the content for the OFF bibliography without change to the content.
* db7b901 moves the content for OFF 7.2.3 without change to the content.
* In beec948, I've started adding content for the start of what needs to go into OFF 5.7.11. Some of this supersedes content currently in the Graphical Primitives / Paints (mostly the same, but with some modification). But most is new content.

This is still in progress: there's more content to be added. Eventually, the current OFF section would be entirely replaced.


PeterCon:
Used wrong GitHub account.

PR #122, opened 2020-12-01
Replace OFF Data types section with more complete section in annex

PeterConstable:


rsheeter:
This one doesn't diff nicely; I really want us to try hard to have good diffs.

PR doesn't appear to remove the original - IMHO it should, I don't think keeping two copies is helpful - and the diff shows it as a pure addition so the changes are hidden. I suggest doing this in two PRs, potentially sending both for review concurrently and noting on the description for the second that it builds upon the first:

1. Reorganize content without making any changes, noting that nothing changed in the description
   * Diff will be large but we don't have to inspect
   * It might be worth doing this once for most or all the content currently under OFF Changes
1. Replace the data types section
   * Now the diff will cleanly indicate what's changing

If you do one big reorg PR then it becomes possible to send numerous small edit PRs concurrently, which I suspect will prove desirable.

PeterConstable:
>PR doesn't appear to remove the original

??
Lines 379 - 385 are the content removed.

rsheeter:
Oh my bad, I thought I'd seen a similar table in the original. Hm, so that maybe presents a new problem: how do we indicate what has changed? Highlight or indicate in the italic text what is added, e.g. _Replace the table defining data types with the following (adding uint24 and Offset24)_?

PeterConstable:
The text is exactly as it needs to be in the OFF amendment. In fact, it's identical to what is in the current working draft.

rsheeter:
Right, but you can't easily tell what's changed when reading. I wonder if we should have a note - not meant to go into the final spec ofc - that says what the change is.

colr-gradients-spec.md
```diff
@@ -1322,6 +1313,29 @@ Changes to the following sections of ISO/IEC 14496-22:2019 Open Font Format (OFF
 - 7.2.3 Item variation stores
 - Bibliography
 
+## A.1 Changes to OFF 4.3 Data types
+
+_Replace the table defining data types with the following:_
```
rsheeter, line 1318: Nit: consider noting the change (add uint24 & Offset24) in some way, otherwise it's not at all obvious what changed

rsheeter (approved):
I think this is fine to merge

PeterConstable:
I'll incorporate a mention of Offset24. (The uint24 type is not new—was added in OT 1.5.)

colr-gradients-spec.md
```diff
@@ -1322,6 +1313,29 @@ Changes to the following sections of ISO/IEC 14496-22:2019 Open Font Format (OFF
...
+| uint16 | 16-bit unsigned integer. |
+| int16 | 16-bit signed integer. |
+| uint24 | 24-bit unsigned integer. |
+| uint32 | 32-bit unsigned integer. |
+| int32 | 32-bit signed integer. |
+| Fixed | 32-bit signed fixed-point number (16.16) |
+| FWORD | int16 that describes a quantity in font design units. |
+| UFWORD | uint16 that describes a quantity in font design units. |
+| F2DOT14 | 16-bit signed fixed number with the low 14 bits of fraction (2.14). |
+| LONGDATETIME | Date and time represented in number of seconds since 12:00 midnight, January 1, 1904. The value is represented as a signed 64-bit integer. |
```
drott, line 1333: At least type LONGDATETIME seems like and addition not related to the COLR additions - why do we add that here?

PeterConstable: I'm not sure what you're comparing against. If you're concerned that it's an addition relative to this document, that's irrelevant. In relation to content to go into OFF, what matters is the diff of the OFF doc. If LONGDATETIME were left out, that would entail _removal_ of LONGDATETIME from the table in OFF clause 4.3. What is in the commit is exactly as it should be.

drott: Yes, I again was coming from a diff-based assumption. True, if we say: "replace the section" then previous types need to stay.

PR #126, opened 2020-12-02
Begin section in annex for OFF 5.7.11 - intro

PeterConstable:
This is the first of several chunks of content that need to go into amendments to 5.7.11. This is all new content, nothing moved from elsewhere in the current doc.


rsheeter:
> This is all new content, nothing moved from elsewhere in the current doc.

That note is much appreciated :)

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
 | Offset24 | 24-bit offset to a table, same as uint24, NULL offset = 0x000000 |
 | Offset32 | Long offset to a table, same as uint32, NULL offset = 0x00000000 |
 
+## A.2 Changes to OFF 5.7.11 - Color Table
+
+_Replace the content of clause 5.7.11 with the following:_
+
+The COLR table adds support for multi-colored glyphs in a manner that is compatible with existing text engines and relatively easy to support with current OpenType font files.
```
rsheeter, line 1207: Optional nit: bit ahead of ourselves to say it's relatively easy (we think so but would others agree?), possibly say "designed to be easy" or some such?

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+The COLR table defines color presentations for glyphs. The color presentation of a glyph is specified as a graphic composition using other glyphs, such as a layered arrangment of glyphs, each with a different color. The term “color glyph” is used informally to refer to such a graphic composition defined in the COLR table; and the term “base glyph” is used to refer to a glyph for which a color glyph is provided.
```
rsheeter, line 1209: > the term “base glyph” is used

I think if I was reading this for the first time I'd be confused here. For me at least an example, like "unicode blah (the emoji whatever) maps to a base glyph which in turn maps to 1..N color glyphs that are composed to produce the final rendering.

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+Processing of the COLR table is done on glyph sequences after text layout processing is completed and prior to final presentation of glyphs. Typically, a base glyph is a glyph that may occur in a sequence that results from the text layout process. In some cases, a base glyph may be a virtual glyph defined within this table as a re-usable color composition.
+
+Two versions of the COLR table are defined.
+
+Version 0 allows for a simple composition of colored elements: a linear sequence of glyphs that are stacked vertically (z-order) as layers. Each layer combines a glyph outline from the &#39;glyf&#39;, CFF or CFF2 table (referenced by glyph ID) with a solid color fill. These capabilities are sufficient to define color glyphs such as illustrated in figure <span style="color:red">5.x</span>.
```
rsheeter, line 1215: Suggest s/z-order/bottom-up z-order/ as used in https://docs.microsoft.com/en-us/typography/opentype/spec/colr.

https://docs.microsoft.com/en-us/typography/opentype/spec/colr doesn't at a glance seem to limit the tables that can provide the glyph. Is there somewhere that says I can't use a bitmap or svg layer in COLR today? - if not let's not add restrictions.

I suggest s/a glyph outline from the 'glyf', CFF or CFF2/a glyph, such as one from glyf or CFF/ to avoid adding rules that may not be in the spec today.

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+![Three emoji glyphs that use layered shapes with solid color fills.](images/colr_v0_emoji_sample.png)
+
+**Figure <span style="color:red">5.x</span> Examples of the graphic capabilities of COLR version 0**
+
+Version 1 supports much richer graphic capabilities. In addition to solid colors, gradient fills can be used, as well as more complex fills using other graphic operations, including affine transformations and diffrent blending modes. Version 1 capabilities allow for color glyphs such as those illustrated in figure <span style="color:red">5.x</span>:
+
+![Three emoji glyphs that use gradient fills and other effects.](images/colr_v1_emoji_sample.png)
+
+**Figure <span style="color:red">5.x</span> Examples of the graphic capabilities of COLR version 0**
```
rsheeter, line 1225: version 1

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+Processing of the COLR table is done on glyph sequences after text layout processing is completed and prior to final presentation of glyphs. Typically, a base glyph is a glyph that may occur in a sequence that results from the text layout process. In some cases, a base glyph may be a virtual glyph defined within this table as a re-usable color composition.
+
+Two versions of the COLR table are defined.
+
+Version 0 allows for a simple composition of colored elements: a linear sequence of glyphs that are stacked vertically (z-order) as layers. Each layer combines a glyph outline from the &#39;glyf&#39;, CFF or CFF2 table (referenced by glyph ID) with a solid color fill. These capabilities are sufficient to define color glyphs such as illustrated in figure <span style="color:red">5.x</span>.
+
+![Three emoji glyphs that use layered shapes with solid color fills.](images/colr_v0_emoji_sample.png)
+
+**Figure <span style="color:red">5.x</span> Examples of the graphic capabilities of COLR version 0**
```
rsheeter, line 1219: consider avoiding inline html, if we ever have to process this to derive something else having it be pure md will make it much easier

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+Version 1 supports much richer graphic capabilities. In addition to solid colors, gradient fills can be used, as well as more complex fills using other graphic operations, including affine transformations and diffrent blending modes. Version 1 capabilities allow for color glyphs such as those illustrated in figure <span style="color:red">5.x</span>:
```
rsheeter, line 1221: typo diffrent

rsheeter: optional, stylistic. "supports much richer" is kind of empty, IMHO it is nice to just directly lean into fact. Version 1 adds gradient fills and blending modes.

rsheeter: Affine transform is a bit out of place, it's more of a size optimization than a new graphic capability. V0 can still draw the transformed shapes, just for higher byte cost.

colr-gradients-spec.md
```diff
@@ -1199,6 +1200,37 @@ _Replace the table defining data types with the following (added row for Offset2
...
+
+![Three emoji glyphs that use gradient fills and other effects.](images/colr_v1_emoji_sample.png)
+
+**Figure <span style="color:red">5.x</span> Examples of the graphic capabilities of COLR version 0**
+
+Version 1 also extends capabilities in variable fonts. A COLR version 0 table can be used in variable fonts with glyph outlines being variable, but no other aspect of the color composition being variable. In version 1, several additional items can be variable, such as the placement of color stops in a gradient or the alpha values applied to colors. The graphic capabilities supported in version 0 and in version 1 are described in more detail below.
```
rsheeter, line 1227: I think it would be fair to say that all new constructs in V1 are deliberately designed to be variable? - we've gone as far as adding new ones like rotate for the sole reason of allowing variation by alternate units.

colr-gradients-spec.md
PeterConstable: I'm certain the intent was that shapes come only from glyph outlines from glyf, CFF or CFF2. There's no way to select bitmaps from EBDT/CBDT, and there'd be no benefit of using bitmaps in in a COLR implementation: you'd have greater control of pixels just using the CBDT table, and trying to integrating with COLR would provide any scaling benefits. Similarly, there's no way to indicate that SVG docs should be used as as sources in a COLR color glyph, and also no benefit from doing so: if SVGs are supported, just use the SVG table directly.

So, while this wasn't previously stated explicitly, it's a defacto reality, and I think we should put a stake in the ground to make this explicit.

(If down the road we want a way to use bitmaps in a COLR color glyph, we'll need to add new formats explicitly, and we can adjust as needed then.)

colr-gradients-spec.md
PeterConstable: My intent was to style these OFF-internal numbering references to call out to WG3 that these require modification. There's no pure-md to add that styling. When it is incorporated, into an OFF draft, all the numbering will get resolved and this can go away.

colr-gradients-spec.md
PeterConstable: >Affine transform is a bit out of place, it's more of a size optimization than a new graphic capability. V0 can still draw the transformed shapes, just for higher byte cost.

Given enough glyphs of very-slightly different shape/size, one could also implement a gradient using V0. Ability to insert a transform into the color glyph definition is certainly an additional capability beyond layers and solid color fills, just as much as using gradient fills or blending modes.

colr-gradients-spec.md
PeterConstable: Yeah, though we need to be careful not to over-state. E.g., composition/blending modes and palette indices aren't in any way variable (because variation of any of those isn't a well-defined concept). I'll try to find balance in revised wording.

colr-gradients-spec.md
```diff
@@ -1204,27 +1204,27 @@ _Replace the table defining data types with the following (added row for Offset2
 
 _Replace the content of clause 5.7.11 with the following:_
 
-The COLR table adds support for multi-colored glyphs in a manner that is compatible with existing text engines and relatively easy to support with current OpenType font files.
+The COLR table adds support for multi-colored glyphs in a manner that integrates with existing text engines rasterizers and that is designed to be easy to support with current OpenType font files.
```
svgeesus, line 1207: perhaps "integrates with the rasterizers of existing text engines" would read better?

colr-gradients-spec.md
```diff
@@ -1204,27 +1204,27 @@ _Replace the table defining data types with the following (added row for Offset2
...
 
-The COLR table defines color presentations for glyphs. The color presentation of a glyph is specified as a graphic composition using other glyphs, such as a layered arrangment of glyphs, each with a different color. The term “color glyph” is used informally to refer to such a graphic composition defined in the COLR table; and the term “base glyph” is used to refer to a glyph for which a color glyph is provided.
+The COLR table defines color presentations for glyphs. The color presentation of a glyph is specified as a graphic composition using other glyphs, such as a layered arrangement of glyphs, each with a different color. The term “color glyph” is used informally to refer to such a graphic composition defined in the COLR table; and the term “base glyph” is used to refer to a glyph for which a color glyph is provided. Processing of the COLR table is done on glyph sequences after text layout processing is completed and prior to final presentation of glyphs. Typically, a base glyph is a glyph that may occur in a sequence that results from the text layout process. In some cases, a base glyph may be a virtual glyph defined within this table as a re-usable color composition.
```
svgeesus, line 1209: Are there any other possibilities, besides layering? If not, why use "such as"

colr-gradients-spec.md
```diff
@@ -1204,27 +1204,27 @@ _Replace the table defining data types with the following (added row for Offset2
...
 
 ![Three emoji glyphs that use layered shapes with solid color fills.](images/colr_v0_emoji_sample.png)
 
 **Figure <span style="color:red">5.x</span> Examples of the graphic capabilities of COLR version 0**
 
-Version 1 supports much richer graphic capabilities. In addition to solid colors, gradient fills can be used, as well as more complex fills using other graphic operations, including affine transformations and diffrent blending modes. Version 1 capabilities allow for color glyphs such as those illustrated in figure <span style="color:red">5.x</span>:
+Version 1 supports additional graphic capabilities. In addition to solid colors, gradient fills can be used, as well as more complex fills using other graphic operations, including affine transformations and different blending modes. Version 1 capabilities allow for color glyphs such as those illustrated in figure <span style="color:red">5.x</span>:
```
svgeesus, line 1221: perhaps "various" blending modes?

svgeesus (commented):
Overall looks great, couple of minor copy-edit suggestions

colr-gradients-spec.md
PeterConstable: Um... it already says, "such as". Is there somewhere else you had in mind?

rsheeter: I interpret "layered arrangement" to informally mean making a stack of layers, as in each layer has a unique z-position. If so, there are indeed other options.For example, we could have two layers that are blended or used to clip one another or w/e. I suspect that might not be something one would typically visualize as "layered"?

rsheeter (approved):

PR #128, opened 2020-12-03
add start of 5.7.11.1 Graphic Compositions

PeterConstable:
All new content, nothing moved. 5.7.11.1 will have a sequence of sub-clauses that will come in subsequent PRs.

This covers some of the same content in the High-level Design and Graphical Primitives / Paints sections of the proposal doc, but I'm not attempting to massage that content into OFF content.


colr-gradients-spec.md
```diff
@@ -1228,6 +1228,60 @@ Version 1 also extends capabilities in variable fonts. A COLR version 0 table ca
...
+
+The graphic compositions in a color glyph definition use a set of 2D graphic concepts and constructs:
+
+* Shapes (or *geometries*)
+* Fills (or *shadings*)
+* Layering—a *z-order*—of elements
+* Composition and blending modes—different ways that the content of a layer is combined with the content of layers above or below it
+* Affine transformations
+
+For both version 0 and version 1, shapes are obtained from glyph outlines in the &#39;glyf&#39;, &#39;CFF &#39; or CFF2 table, referenced by glyph ID. Colors used in fills are obtained from the CPAL table.
```
rsheeter, line 1241: Both here and above in "Each layer combines a glyph outline from the 'glyf', CFF or CFF2 table" I don't think we should word this to be restrictive. A "such as" might be appropriate.

colr-gradients-spec.md
```diff
@@ -1228,6 +1228,60 @@ Version 1 also extends capabilities in variable fonts. A COLR version 0 table ca
...
+
+**Figure 5.8</span> Basic graphic capabilities of COLR version 0**
+
+The basic concepts also apply to color glyphs defined using the version 1 formats: shapes are arranged in layers and have fills. But the additional formats of version 1 support much richer capabilities. In a version 1 color glyph, graphic constructs and capabilities are represented primarily in *Paint* tables, which are linked together in a *directed, acyclic graph*. Several different Paint formats are defined, each describing a particular type of graphic operation:
+
+* A PaintColrLayers table provides a layering structure used for creating a color glyph from layered elements. A PaintColrLayers table can be used at the root of the graph, providing a base layering structure for the entire color glyph definition. A PaintColrLayers table can also be nested within the graph, providing a set of layers to define some graphic sub-component within the color glyph.
+
+* The PaintSolid, PaintLinearGradient, and PaintRadialGradient tables provide basic fills, using color entries from the CPAL table.
+
+* The PaintGlyph table provides glyph outlines as the basic shapes.
```
rsheeter, line 1257: I think it's important to mention that PaintGlyph paints the region of some other paint bounded by a glyph outline. Potentially could refer to figure 5.8 to give a quick example: the triangle can be a PaintGlyph with the triangle outline (which has no intrinsic color) and a PaintSolid

colr-gradients-spec.md
```diff
@@ -1228,6 +1228,60 @@ Version 1 also extends capabilities in variable fonts. A COLR version 0 table ca
...
+
+* The PaintComposite table supports alternate composition and blending modes for two sub-graphs.
+
+* The PaintColrGlyph table allows a color glyph definition, referenced by a base glyph ID, to be re-used as a sub-graph within multiple color glyphs.
+
+In a simple color glyph description, a PaintGlyph table might be linked to a PaintSolid table, for example, representing a glyph outline filled using a basic solid color fill. But the PaintGlyph table could instead be linked to a much more complex sub-graph of Paint tables, representing a shape that gets filled using the more-complex set of operations described by the sub-graph of Paint tables.
+
+The graphic capabilities are described in more detail in 5.7.11.1.1 – 5.7.11.1.7. The formats used for each are specified 5.7.11.2.
+
+**5.7.11.1.1 Colors and solid color fills**
```
rsheeter, line 1269: Would definitely like to see these be populated by moving existing content, not by purely writing new content.

rsheeter (approved):

colr-gradients-spec.md
PeterConstable: I disagree with leaving open a suggestion that bitmaps from CBDT/EBDT or the bitmaps that result from rendering SVG docs from the SVG table can be used as shapes in COLR: the interaction with scaling is completely unspecified, hence there's no way to determine conformance. The COLR table was never intended to be used in combination with CBDT/EBDT or SVG,

colr-gradients-spec.md
PeterConstable: This is an outline. The details you seek will come in 5.7.11.1.3.

PR #135, opened 2020-12-05
revision of color line section

PeterConstable:
This is a significant revision of the color line section, with a clearer, more accurate and more complete description.

Of the PRs to come, this will probably be the biggest revision of existing content.


drott:
I'll need some more time to review this and will try to get to this soon.

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
```
drott, line 1040: Generally, please try to line-break the markdown to 80 columns if possible.

drott, line 1040: Generally, please try to line-break the markdown to 80 columns if possible.

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
 
-Color stops are defined at color stop positions. Color stop position 0 maps to
-the start point of a linear gradient or the center of the first circle of a
-radial gradient. Color stop position 1 maps to the end point of a linear
-gradient or the center of the second circle of a radial gradient.  In the
-interval [0, 1] the color line must contain at least one color stop, but may
-contain multiple color stops that define the gradient.
+A color line is not a line in a geometric sense: on its own, a color line has no positioning, orientation or size within a design grid. These geometric aspects are part of the definition of a linear or radial gradient. Specifically, the gradient definition will specify positions in the design grid that correspond to the real values 0 and 1 in the color line, with placement in the design grid for other numeric values of the color line interpolated from the placement of 0 and 1.
```
drott, line 1045: As is, contrasting it to a geometric line, I find this to be rather having the character of a note. Perhaps we can rephrase positively:

```suggestion
On its own, a color line has no positioning, orientation or size within a design grid. By mapping and referencing a color line, linear and radial gradient definitions will specify positions in the design grid that correspond to the real values 0 and 1 in the color line, as well as rules for where to draw interpolated colors of the color line interpolated that follow from the placement of color stops 0 and 1.
```

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
 
-Outside the defined interval, the gradient pattern in between the outer defined
-positions is repeated according to the color line [extend
-mode](#extend-mode).
+A color stop is defined by a real number, the *stop offset*, and a color. A color line is defined with at least one color stop within the interval [0, 1]. Additional color stops can be specified within or outside the interval [0, 1]. (Stop offsets are represented using F2DOT14 values, therefore color stops can only be specified within the range [-2, 2). See <span style="color:red">5.7.11.2.x</span> for format details.) If only one color stop is specified, that color is used for the entire color line; at least two color stops are needed to create color gradation.
```
drott, line 1047: > A color line is defined with at least one color stop within the interval [0, 1].

Filed #148 for that, I don't think we need to enforce that, but we can handle that separately.

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
+Suppose a color line has two or more color stops, and consider these to be ordered in increasing stop offset value. Color gradation is defined over the interval from the first color stop, through the successive color stops, to the last color stop. Between consecutive color stops, color values are interpolated between the colors of the two stops.
 
-Limiting the specified interval to a sub-range of [0, 1] allows for looping
-through colors repeatedly along the mapped distance, without having to encode
-them multiple times.  In that sense, our color line is similar to CSS
-[repeating-linear-gradient()](https://developer.mozilla.org/en-US/docs/Web/CSS/repeating-linear-gradient)
-and
-[repeating-radial-gradient()](https://developer.mozilla.org/en-US/docs/Web/CSS/repeating-radial-gradient)
-functions.
+> **_TBD: Does interpolation of colors need further specification?_**
```
drott, line 1051: Minor: I prefer to file issues instead of inline TODOs.

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
 
-If there are multiple color stops defined for the same coordinate, the first one
-is used for computing the color value for values below the coordinate, the last
-one is used for computing the color value for values above. All other color
-stops for this coordinate are ignored.
+Suppose a color line has two or more color stops, and consider these to be ordered in increasing stop offset value. Color gradation is defined over the interval from the first color stop, through the successive color stops, to the last color stop. Between consecutive color stops, color values are interpolated between the colors of the two stops.
```
drott, line 1049: Can we generalize this and avoid the specific "Suppose...", to e.g.:
Between adjacent color stops, at any position between the two stops, the color is the linearly interpolated color value at that position.

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
-  and
-  [radial-gradient()](https://developer.mozilla.org/en-US/docs/Web/CSS/radial-gradient)
-  functions.
+While the color gradation is specified over a defined interval, the color line continues indefinitely outside that interval in both directions. The color pattern outside the defined interval is repeated according to the color line’s *extend mode*. Three extend modes are supported:
 
-* a repeated gradient along the gradient positions (linear or radial): divide 1
-  by the number of desired repetitions and use the result as your maximum color
-  stop, then use color line extend mode *repeat* to have it continue outside the
-  defined interval.
+* Pad: outside the defined interval, the color of the closest color stop is used. Using a string as an analogy, given a sequence “ABC”, it is extended to “…*AA* ABC *CC*…”.
```
drott, line 1057: Formatting suggestion, and an optional suggestion to use less programming-specific terminology:

```suggestion
* **Pad:** outside the defined interval, the color of the closest color stop is used. Using a sequence of letters as an analogy, given a sequence “ABC”, it is extended to “…*AA* ABC *CC*…”.
```

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
 
-* a mirrored / color-circle gradient: divide 1 by two times the number of
-  desired full color stripes, and define the color stops between the 0 and the
-  result of this division, then use color line extend mode *reflect* to have it
-  continue mirrored.
+* Repeat: The color line is repeated over repeated multiples of the defined interval. For example, if color stops are specified for defined interval of [0, 0.7], then the pattern is repeated above the defined interval for intervals [0.7, 1.4], [1.4, 2.1], etc.; and also repeated below the defined interval for intervalues [-0.7, 0], [-1.4, -0.7], etc. In each repeated invterval, the first color is that of the farthest defined color stop. By analogy, given a sequence “ABC”, it is extended to “…*ABC* ABC *ABC*…”.
 
-![Repeating linear gradient](images/repeating_linear.png) ![Repeating radial gradient](images/repeating_radial.png)
+* Reflect: The color line is repeated over repeated intervals, as for the repeat mode. However, in each repeated interval, the ordering of color stops is the reverse of the adjacent interval. By analogy, given a sequence “ABC”, it is extended to “…*ABC CBA* ABC *CBA ABC*…”.
```
drott, line 1061:
```suggestion
* **Reflect:** The color line is repeated over repeated intervals, as for the repeat mode. However, in each repeated interval, the ordering of color stops is the reverse of the adjacent interval. By analogy, given a sequence “ABC”, it is extended to “…*ABC CBA* ABC *CBA ABC*…”.
```

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
-  by the number of desired repetitions and use the result as your maximum color
-  stop, then use color line extend mode *repeat* to have it continue outside the
-  defined interval.
+* Pad: outside the defined interval, the color of the closest color stop is used. Using a string as an analogy, given a sequence “ABC”, it is extended to “…*AA* ABC *CC*…”.
 
-* a mirrored / color-circle gradient: divide 1 by two times the number of
-  desired full color stripes, and define the color stops between the 0 and the
-  result of this division, then use color line extend mode *reflect* to have it
-  continue mirrored.
+* Repeat: The color line is repeated over repeated multiples of the defined interval. For example, if color stops are specified for defined interval of [0, 0.7], then the pattern is repeated above the defined interval for intervals [0.7, 1.4], [1.4, 2.1], etc.; and also repeated below the defined interval for intervalues [-0.7, 0], [-1.4, -0.7], etc. In each repeated invterval, the first color is that of the farthest defined color stop. By analogy, given a sequence “ABC”, it is extended to “…*ABC* ABC *ABC*…”.
```
drott, line 1059: Formatting, typos.

```suggestion
* **Repeat:** The color line is repeated over repeated multiples of the defined interval. For example, if color stops are specified for defined interval of [0, 0.7], then the pattern is repeated above the defined interval for intervals [0.7, 1.4], [1.4, 2.1], etc.; and also repeated below the defined interval for intervals [-0.7, 0], [-1.4, -0.7], etc. In each repeated interval, the first color is that of the farthest defined color stop. By analogy, given a sequence “ABC”, it is extended to “…*ABC* ABC *ABC*…”.
```

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,41 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
-* a repeated gradient along the gradient positions (linear or radial): divide 1
-  by the number of desired repetitions and use the result as your maximum color
-  stop, then use color line extend mode *repeat* to have it continue outside the
-  defined interval.
+* Pad: outside the defined interval, the color of the closest color stop is used. Using a string as an analogy, given a sequence “ABC”, it is extended to “…*AA* ABC *CC*…”.
 
-* a mirrored / color-circle gradient: divide 1 by two times the number of
-  desired full color stripes, and define the color stops between the 0 and the
-  result of this division, then use color line extend mode *reflect* to have it
-  continue mirrored.
```
drott, line 1092: This type of cookbook section is not contained in the new text and I find it helpful to keep as it illustrates the utility of defining ranges for sub ranges of [0,1]. Can we add it back as a suitable subsection?

drott (commented):
Thanks for adding detail to this section and the additional images for the extend modes. Some comments above/below.

colr-gradients-spec.md
PeterConstable: Agree we should have issues, but I think it's helpful to call out in the draft text so that the open issues are less likely to be overlooked.

colr-gradients-spec.md
PeterConstable: The wording has a lot of assumptions and wasn't clear. But I'll work on clarifying what I think was the intent.

colr-gradients-spec.md
PeterConstable: Opened #150

colr-gradients-spec.md
```diff
@@ -1040,84 +1040,94 @@ COLR version 1 supports two types of gradients: linear gradients, and radial gra
...
-
-For numbers outside the defined interval, the color continues to map as if the
-interval would continue mirrored from the previous interval. This allows
-defining stripes in rotating colors.
+A color line is a function that maps real numbers to color values to define a
+one-dimensional gradation of colors, to be used in the definition of linear or
+radial gradients. A color line is defined as a set of one or more color stops,
+each of which maps a particular real number to a specific color.
+
+On its own, a color has no positioning, orientation or size within a design
```
drott, line 1048:
```suggestion
On its own, a color line has no positioning, orientation or size within a design
```

I think `line` i missing here, otherwise the paragraph reads well now, thanks.

colr-gradients-spec.md
drott: Thanks for opening the issue - one reason I find inline TBDs somewhat redundant is because we may also find in the discussion of the issue that we do not need further clarifications. Then we can just close the issue. But for removing the inline TBDs we'd need another edit. My preference would still be to remove them inline. @rsheeter, @anthrotype - opinions?

colr-gradients-spec.md
drott: Thank you for adding this section back in. I find it not helpful to read blanket judgements applied like "wasn't clear" / "had a lot of assumptions" without stating what exactly was not clear or which assumptions should have been called out explicitly. Similarly, in my opinion a commit description stating "with a clearer, more accurate and more complete description." is broad statement potentially not shared by all readers. Our perceptions about the clarity, precision or meaning of a text can individually differ. 

In this case of the cookbook section, I personally found the the previous text just as clear as the new one. I would prefer that we talk about what specifically did not seem clear and then discuss, iterate the wording and improve it so we arrive at spec text that is most broadly understood.

drott (approved):
LGTM with word `line` inserted, and still with a preference for removing inline TBDs.

PeterConstable:
@drott 

>Thank you for adding this section back in. I find it not helpful to read blanket judgements applied like "wasn't clear" / "had a lot of assumptions" without stating what exactly was not clear or which assumptions should have been called out explicitly. Similarly, in my opinion a commit description stating "with a clearer, more accurate and more complete description." is broad statement potentially not shared by all readers. Our perceptions about the clarity, precision or meaning of a text can individually differ.

Fair critique. Thanks.

.In this case of the cookbook section, I personally found the the previous text just as clear as the new one. I would prefer that we talk about what specifically did not seem clear and then discuss, iterate the wording and improve it so we arrive at spec text that is most broadly understood.

When I first read the original wording, it seemed like it was assuming that gradients would be painted only for the range [0,1]. But of course the gradients are unbounded: so, if you want 5 repetitions (say), you can define stops from 0 to 1, use the repeat extend mode, and you'll get an unlimited number of repetitions. 

>In order to achieve a repeated gradient along the gradient positions (linear or radial): divide 1
  by the number of desired repetitions...

To achieve repetitions, it doesn't matter where you set the stops; just set the repeat extend mode. This seemed pointless at best, and potentially misleading, which is why I dropped it.

When you said you thought it was helpful to keep, i read once more to see what unstated assumptions there might be that would make it useful — I assumed if you thought it was useful then it must, in fact, be useful. It occurred to me that the guidance would be useful _if you want to align the pattern **_geometrically_**. That wasn't stated originally.

PR #137, opened 2020-12-05
Revise radial gradient section

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
 
 **5.7.11.1.2.3 Radial gradients**
 
-Radial gradients in this proposal are defined based on circles. If subject to
-a transform (via `PaintTransformed`) those circles may become ellipses.
+A basic radial gradient provides gradation of colors that radiates in all directions from an origin. The radial gradients supported in COLR version 1 provide a superset of this functionality: color gradation is defined along a cylinder defined by two circles. In the general case, the circles can have different radii to create a cone. A basic radial gradient is formed by one circle having a radius of zero with center located inside the other circle.
```
drott, line 1167: See below, for brevity and avoiding building definitions such as "basic radial gradient" that are not strictly needed for this spec, I would suggest to move the third paragraph up here and either strip this section or move it to a note and shorten it.

colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
...
-color at color line position 0. The circle c₁, r₁ will be drawn with the color
-at color line colorLine position 1.
+NOTE: Some graphics libraries or other specifications support “radial” gradients with the same capabilities described here, whereas in other contexts, a “radial” gradient might provide more limited capabilities. In some contexts, the type of gradient defined here is referred to as a “two point conical” gradient.
 
-The drawing algorithm radial gradients follows the [HTML WHATWG Canvas spec for
-createRadialGradient()](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-createradialgradient).
-Quoting and adapting from there.  With circle center points c₀ and c₁ defined as
-c₀ = (x₀, y₀) and c₁ = (x₁, y₁):
+A radial gradient is defined by two circles, one with center c₀ and radius r₀, and another with center c₁ and radius r₁, plus a color line. The color line is related to the two circles by aligning stop offset 0 with a point on the perimeriter of the first circle (with center c₀) and aligning stop offset 1.0 with a corresponding point (the same angular direction from the center) on the perimeter of the second circle.
```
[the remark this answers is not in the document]
drott: I assume you did want to address the fact that depending on repeat mode, not the full circle is drawn with the color at the respective stop offset, is that the case?

It's good that you fixed and removed the remaining notion of `optionally transformed`.

I don't think the "alignment of stop offsets to a point on the perimeter of..." formulation adds precision here as the algorithm describes the pixel and color placement accurately. We only and mainly need to specify the direction or association of the color line to circles, I suggest instead:

```suggestion
A radial gradient is defined by two circles, one with center c₀ and radius r₀, and another with center c₁ and radius r₁, plus a color line. The color line is mapped to the two circles by associating stop offset 0 with first circle (with center c₀) and associating stop offset 1.0 with the second circle (with center c₁), with rules for which pixels to draw as in the algorithm below.
```

colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
...
    Let y(ω) = (y₁-y₀)ω + y₀<br>
    Let r(ω) = (r₁-r₀)ω + r₀<br>
-   Let the color at ω be the color at that position on the gradient color line
-   (with the colors coming from the interpolation and extrapolation described above).
+   Let the color at ω be the color at that position on the gradient color line.
 3. For all values of ω where r(ω) > 0, starting with the value of ω nearest to
    positive infinity and ending with the value of ω nearest to negative
    infinity, draw the circumference of the ellipse resulting from translating
-   circle with radius r(ω) by affine transform at position (x(ω), y(ω)), with
+   a circle with radius r(ω) by affine transform such that the center is at position (x(ω), y(ω)), with
```
drott, line 1189:
```suggestion
   a circle with radius r(ω) centered at position (x(ω), y(ω)), with
```

Suggest to strip notion of transform. I think that's also a leftover from when we had a transform parameter here.

colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
...
    the color at ω, but only painting on the parts of the bitmap that have not
    yet been painted on by earlier circles in this step for this rendering of the
    gradient.
 
-![Example radial gradient rendering](images/radial_gradients.png)
+Figure 5.13 illustrates a radial gradient using the three different color line extend modes. The color line is defined with stops for the interval [0, 1]. Note that the circles that define the gradient are not stroked as part of the gradient itself. Stroked circles have been overlaid in the figure to illustrate the color line and the region that is painted in relation to the two circles.
```
drott, line 1194: This paragraph, or the Figure caption now lacks a description of the colors on the color stops, that the previous text had: 

"color stops for blue at 0, yellow at 0.5 and red at 1."

colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
...
-generated from <a href="images/radial_gradient_direction.svg">images/radial_gradient_direction.svg</a>*
+**Figure 5.14 Radial gradients with start and end circles swapped.**
 
-*__Note:__ It follows from the algorithm that: a) When both radii are 0, r₀ = r₁
-= 0, then r(ω) is always 0 and nothing is painted. b) If the centers of the
-circles are distinct, one has a zero radius and lies outside the radius of the
-other, the other has a non-zero radius, the resulting shape resembles a cone
-that is open to one side, see Figure 3 below. c) If the circles centers are
-identical, one has a zero radius, one has a non-zero radius, then the gradient
-is well-defined and is painted according to the algorithm.*
```
drott, line 1234: It seems to me that these clarifications are dropped in the new text, but they were added to address the questions raised in an issue what happens in edge cases. I suggest to add those back in.

colr-gradients-spec.md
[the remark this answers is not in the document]
PeterConstable: >I think it makes sense to use this as the first paragraph of this section instead of starting with a comparison and definition of what a "basic radial gradient" is intended to mean. We do not need to go into details what other spec, implementations or documentation calls their gradients. For brevity and conciseness, I think we should focus on what we define a radial gradient to mean in this spec.

Because the term "radial gradient" gets used in different contexts in different ways, I think there should be _some_ clarification to distinguish from other uses of the term. But it can be more concise.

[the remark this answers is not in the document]
PeterConstable: >I assume you did want to address the fact that depending on repeat mode, not the full circle is drawn with the color at the respective stop offset, is that the case?

Not sure what you have in mind. Step 3 of the algorithm does say,

>but only painting on the parts of the bitmap that have not yet been painted on by earlier circles

Do you think this should be called out in prose?

colr-gradients-spec.md
PeterConstable: The existing text didn't completely cover all cases (e.g., b is overly specific, and another important edge case is not mentioned). I'll re-incorporate revised text.

colr-gradients-spec.md
```diff
@@ -1164,90 +1164,72 @@ glMatrix.js to work)*
...
-a transform (via `PaintTransformed`) those circles may become ellipses.
+A basic radial gradient provides gradation of colors that radiates in all directions from an origin. The radial gradients supported in COLR version 1 provide a superset of this functionality: color gradation is defined along a cylinder defined by two circles. In the general case, the circles can have different radii to create a cone. A basic radial gradient is formed by one circle having a radius of zero with center located inside the other circle.
 
-A radial gradient in this proposal is a gradient between two—optionally
-transformed—circles, namely with center c₀ and radius r₀, and center c₁ and
-radius r₁ and a specified color line.  The circle c₀, r₀ will be drawn with the
-color at color line position 0. The circle c₁, r₁ will be drawn with the color
-at color line colorLine position 1.
+NOTE: Some graphics libraries or other specifications support “radial” gradients with the same capabilities described here, whereas in other contexts, a “radial” gradient might provide more limited capabilities. In some contexts, the type of gradient defined here is referred to as a “two point conical” gradient.
```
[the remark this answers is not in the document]
PeterConstable: See earlier reply: I think _something_ needs to be kept.

colr-gradients-spec.md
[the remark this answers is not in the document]
PeterConstable: Resolved comments except for the one mentioned in my last reply.

Lorp:
To aid understanding, I would recommend that what I expect to be the normal use of radial gradient (one circle enclosed in the other, as in current Fig. 5) be placed first, i.e. before Fig. 3 and Fig. 4. I would also recommend that the first example is a simple gradient with two stops, only later proceeding to 3+ stops and repeat, etc.

It also seems to me that the third item in Fig. 4 is erroneous, and can be dropped. Perhaps I am not understanding Fig. 4 correctly.

PeterConstable:
> It also seems to me that the third item in Fig. 4 is erroneous, and can be dropped. Perhaps I am not understanding Fig. 4 correctly.

The third item in Fig 5.15 (was fig 4) is correct. The first item matches 5.13; the second swaps the circles, but keeps the color line the same. The third also reverses the color line (so that red is on the left, as in the first item).

PeterConstable:
> To aid understanding, I would recommend that what I expect to be the normal use of radial gradient (one circle enclosed in the other, as in current Fig. 5) be placed first, i.e. before Fig. 3 and Fig. 4. 

I think this touches on my concern that the most familiar notion of "radial gradient" for many would not be what _this_ radial gradient can do. But, if you reverse the figures, I don't think the case shown in the last figure gives the best illustration of the the capabilities of the algorithm.

> I would also recommend that the first example is a simple gradient with two stops, only later proceeding to 3+ stops and repeat, etc.

I think having a third stop is important for explaining the extend modes. Else, the repeat and reflect modes will be less distinct. And I wouldn't want to have figures for 2 stops and separate figures for 3 stops, at least certainly not in the radial gradients section. (Maybe an additional example of color lines would be OK in the color line section.)

Lorp:
Ah, sorry I missed the text about colour line reversal. I think it may be confusing and unnecessary to talk about colour line reversal, however. It‘s nice to have some variables stable throughout.

PeterConstable:
> 
> 
> Ah, sorry I missed the text about colour line reversal. I think it may be confusing and unnecessary to talk about colour line reversal, however. It‘s nice to have some variables stable throughout.

Hmmm... Interesting to hear your reaction. My initial impulse was along that line, but it was to show the first and third versions, to keep the appearance as similar as possible—the only change being the bend of the colour arcs. That was the first version of this figure that I provided. Dominik expected the opposite: only change the circles, which made two things in the appearance change. So, the compromise I came up with was to show all three with single code changes from first to second, and from second to third.

So, imagine you only had the first and second: would the main point of the example—the direction of the bend of the colour arcs, be more clear? Less clear? Same?

Lorp:
The colour line is an easy concept to grasp, but the patterns from various circle placements are harder. The flip in the colour line therefore distracted me from understanding the arcs. Also, can you think of a better term than “bends towards x”? If it was a sheet of plastic, you’d “bend it away from x”.

PeterConstable:
> The colour line is an easy concept to grasp, but the patterns from various circle placements are harder. The flip in the colour line therefore distracted me from understanding the arcs. 

So, if the figure only showed the first and second gradients, would that make it more clear?

>Also, can you think of a better term than “bends towards x”? If it was a sheet of plastic, you’d “bend it away from x”.

Yeah, I wondered about that. Perhaps this?

> when the two circles are not overlapping, the arc of color is concave from the perspective of circle 0, convex from the perspective of circle 1.

Got another suggestion?

Lorp:
> when the two circles are not overlapping, the arc of color is concave from the perspective of circle 0

That seems to be inverted! It’s also complicated by repeat and reflect modes.

How about this?

> when the two circles are not overlapping, and ignoring the effect of repeat and reflect modes, the arcs of constant color are centered on circle 1 and are thus convex from the perspective of the center of circle 0.

PeterConstable:
> How about this?
> 
> > when the two circles are not overlapping, and ignoring the effect of repeat and reflect modes, the arcs of constant color are centered on circle 1 and are thus convex from the perspective of the center of circle 0.

I think your and my understanding of "convex from the perspective of" are different. But "centered on the center of circle 1" seems clear and unambiguous.

Lorp:
> NOTE: … if one circle is entirely contained within the other: in that case, the individual colors are complete, concentric circles

They are not concentric. Suggest delete that word.

PeterConstable:
>  "centered on the center of circle 1" seems clear and unambiguous.

I've revised the wording, and also revised the caption for figure 5.15 to focus on the arc, not swapping of circles.

Lorp:
If object A is deep inside the earth and object B is in orbit around the earth, then the earth is concave with respect to object A and convex with respect to object B.

PeterConstable:
> 
> 
> > NOTE: … if one circle is entirely contained within the other: in that case, the individual colors are complete, concentric circles
> 
> They are not concentric. Suggest: “… the arcs of constant color are complete circles”.

Right! Good catch. Updated.

colr-gradients-spec.md
```diff
@@ -1135,90 +1135,87 @@ Figure 5.12 illustrates linear gradients using the three different color line ex
...
+* If the centers of the circles are distinct but the radii are the same, and neither circle is contained within the other, then the result will be a strip, similar to the flattened projection of a circular cylinder. The surface outside the strip is not painted. (See figure 5.14.)
+* If the radii of the circles are different but one circle is entirely contained within the radius of the other circle, the gradient will radiate outward in all directions from the inner circle, and the entire surface will be painted. (See figure 5.16.)
+
+NOTE: In the case of one circle being entirely contained within the other, if the radius of the inner circle is zero, then the gradient will radiate from the center point. No extension of the color line beyond that circle will be painted. Thus, the loop of the algorithm can be shortened to exclude values of ω corresponding to the inner circle (either 0 or 1) or beyond.
```
drott, line 1170: As this applies more generally (not just for when circles are contained within each other) but for whenever the radius becomes zero and the algorithm already states "...for ω where r(ω) > 0" this seems clear to me without this note.

colr-gradients-spec.md
```diff
@@ -1135,90 +1135,87 @@ Figure 5.12 illustrates linear gradients using the three different color line ex
...
+
+> **_TBD: What should the expected behaviour be in the case of both circles exactly overlapping with r > 0? (Only the extensions get painted, but how?)_**
```
drott, line 1172: Minor and optional: instead of TBDs in the text, let's keep track of TODOs in issues rather than in text.

colr-gradients-spec.md
```diff
@@ -1135,90 +1135,87 @@ Figure 5.12 illustrates linear gradients using the three different color line ex
...
+This is illustrated in figure 5.15, in which three radial gradients are shown.
+The first is the same as the first gradient shown in figure 5.13, using the pad
+extend mode. In this first gradient, circle 0 is the small circle, on the left. In
+the second gradient of figure 5.15, the start and end circles are reversed: circle
 0 is the large circle, on the right. The color line is kept the same, and so the
 red end starts at circle 0, now on the right. In the third gradient, the order
 of stops in the color line is also reversed to put red on the left.
+The key difference to notice between the gradients in figure 5.15 is the way colors
+are painted in the interior: when the two circles are not overlapping, the arcs
+of constant color are centered on circle 1.
```
drott, line 1200: Perhaps
```suggestion
of constant color are concentric to circle 1.
```
?

colr-gradients-spec.md
```diff
@@ -1135,90 +1135,87 @@ Figure 5.12 illustrates linear gradients using the three different color line ex
...
-![Radial gradients with one circle contained within the other](images/radial_gradients_circle_within_circle.png)
+**Figure 5.16 Radial gradients with one circle contained within the other.**
 
-*__Figure 5:__ Radial gradients with one circle contained within the other. (Illustration generated from [images/radial_gradients_circle_within_circle.svg](images/radial_gradients_circle_within_circle.svg).)*
+NOTE: A scale transformation (see 5.7.11.1.5) can flatten shapes to resemble lines. If a radial gradient is nested in the child sub-graph of a transformation that flattens the circles so that they are nearly lines, the centers may still be separated by some distance. In that case, a radial gradient would appear as a strip or a cone filled with a linear gradient.
 
-*__Note:__ When a radial gradient is nested below a transformation which flattens
-the circles so that they resemble lines, a radial gradient may appear as a
-strip or a cone filled with a linear gradient.*
+> **_TBD: We still need to specify required behaviour for the case in which the transform really flattens the two circles, and the centers, to a line._**
```
drott, line 1218: See above, suggest to track TBDs and TODOs in issues, rather than inline.

PeterConstable: Opened #150

colr-gradients-spec.md
PeterConstable: No: I suggested concentric earlier, but as @Lorp rightly pointed out, the circles are not in general concentric. (The interpolated circles will only be concentric if c0 = c1.)

colr-gradients-spec.md
PeterConstable: Opened #150

colr-gradients-spec.md
PeterConstable: 'Centered on circle 1" still doesn't seem quite right, though, so have revised wording.

drott (approved):

PR #139, opened 2020-12-05
Add content for Layers section

PeterConstable:
Entirely new content.


colr-gradients-spec.md
```diff
@@ -1224,6 +1224,32 @@ strip or a cone filled with a linear gradient.*
...
+
+Figure 5.18 illustrates layers using version 0 formats.
+
+![Version 0: Color glyphs are defined by slices of a layer records array.](images/colr_layers_v0.png)
+
+**Figure 5.18 Version 0: Color glyphs are defined by slices of a layer records array.**
+
+The version 1 formats also define layer sets as slices in an array. In this case, however, the array is an array of offsets to paint tables, and is contained in a LayerV1List table. Each referenced paint table is the root of a graph of paint tables that specifies a graphic composition to be used as a layer. Within a given slice, the first offset provides the content for the bottom layer, and each subsequent offset provides content that overlays the preceding content. Definition of the slices (the sub-sequence of offsets for each slice) are give in PaintColorLayers tables.
```
rsheeter, line 1237: "The version 1 formats" => "The version 1 format"

rsheeter, line 1237: This isn't quite right; the list serves a different purpose in V1 and is optional.

COLR v1 defines layer sets as a DAG. Paints _may_ be written into a layer list but do not *have* to be. A COLR v1 font can exist with 0 entries in the LayerV1List.

For example, it would be entirely valid to have a BaseGlyphV1Record with an offset to a PaintGlyph which in turn referenced a PaintSolid and a valid glyph id with no entries in LayerV1List. Pretty sure I even have a unit test to confirm FontTools can output this type of thing.

The reason to write down Paint offsets in the LayerV1List is to allow them to be accessed via PaintColrLayers, thus permitting a compiler to exploit reusable subsequences.

There is a note to this effect buried in https://github.com/googlefonts/colr-gradients-spec#colr-v1-data-structures, LayerV1List table: "Only layers referenced by PaintColrLayers (format 1) records need to be encoded here." BaseGlyphV1Record similarly notes "Offset to Paint, typically a PaintColrLayers."

colr-gradients-spec.md
```diff
@@ -1224,6 +1224,32 @@ strip or a cone filled with a linear gradient.*
...
+
+![Version 1: PaintColrLayers tables specify slices within the LayerV1List, providing a layering of content defined in sub-graphs.](images/colr_layers_v1.png)
+
+**Figure 5.19 Version 1: PaintColrLayers tables specify slices within the LayerV1List, providing a layering of content defined in sub-graphs.**
+
+A PaintColorLayers table is typically used as the root of a color glyph definition, providing a base layering structure for the color glyph. In this usage, the PaintColrLayers table is referenced by a BaseGlyphV1Record, which specifies the root of the graph of a color glyph definition for a given base glyph. This is illustrated in figure 5.20.
+
+![PaintColrLayers table used as the root of a color glyph definition.](images/colr_PaintColrLayers_as_root.png)
+
+**Figure 5.20 PaintColrLayers table used as the root of a color glyph definition.**
```
rsheeter, line 1249: Strongly recommend addition of an example where the BaseGlyphV1Record does *not* use PaintColrLayers.

colr-gradients-spec.md
```diff
@@ -1224,6 +1224,32 @@ strip or a cone filled with a linear gradient.*
...
+
+A PaintColorLayers table can also be nested more deeply within the graph, providing a layer structure to define some component within a larger color glyph definition. The ability to nest a PaintColrLayers table within a graph creates the potential to introduce a cycle within the graph. This is not valid, however: graphs shall be acyclic.
```
rsheeter, line 1251: graphs shall not be acyclic feels like it should be a reference to the section that suggests how this should be accomplished?

rsheeter:
Loving the examples, just need to tweak the explanation of the LayerV1List. I tried to explain but please kick me in the shins if I failed miserably :)

rsheeter (changes requested):
Apparently I must comment here. The LayerV1List description needs adjustment.

colr-gradients-spec.md
PeterConstable: The v1 formats in question are the combination of LayerV1List and PaintColrLayers. So, "formats" (plural) was intended.

PeterConstable: Well, I think what is stated there is factually correct in what it says. But perhaps it could be misleading by the comparison with the v0 formats since, in v0, the layers are _the (only) way_ in which a color glyph is defined, whereas in v1, as you point out, the layers are optional. I agree that should be made clear.

colr-gradients-spec.md
PeterConstable: Agreed.

colr-gradients-spec.md
PeterConstable: Agreed. Will figure out later where to put that and reference.

PeterConstable: #156

colr-gradients-spec.md
```diff
@@ -1246,6 +1246,44 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
 
 **5.7.11.1.4 Layers**
 
+Layering of visual elements was introduced above, in the introduction to 5.7.11.1. Both version 0 and version 1 support layers, though in different ways. In version 0, layers are fundamental: they are the sole way in which separate elements are composed into a color glyph. When using version 1 formats, however, use of layering is optional.
```
rsheeter, line 1249: I think this is confusing two things: 1) the existence of a vector of all layers (not strictly required in v1) 2) whether you have layering in the sense of a glyph drawn of z-ordered content (generally speaking true of both [subject to some nuance around PaintComposite]). I suggest replacing with something like:

Layering of visual elements was introduced above, in the introduction to 5.7.11.1. Both version 0 and version 1 support layers. In version 0 there is a single vector of all layers, referenced by all glyphs. In version 1 there is a vector of layers but it's use is optional, required only for sharing of layers between color glyphs using PaintColrLayers.

colr-gradients-spec.md
```diff
@@ -1246,6 +1246,44 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+
+For version 0, an array of LayerRecords is created. Each LayerRecord specifies a glyph ID and a CPAL entry (a shape and solid color fill). Each color glyph definition is a slice from that array (that is, a contiguous sub-sequence), specified in a BaseGlyphRecord for a particular base glyph. Within a given slice, the first record specifies the content of the bottom layer, and each subsequent record specifies content that overlays the preceding content(increasing z-order). A single array is used for defining all color glyphs. The LayerRecord slices for two base glyphs may overlap, though often will not overlap.
```
rsheeter, line 1251: Given the edit suggested above I think you can delete "A single array is used for defining all color glyphs."

colr-gradients-spec.md
```diff
@@ -1246,6 +1246,44 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+
+Figure 5.18 illustrates layers using version 0 formats.
+
+![Version 0: Color glyphs are defined by slices of a layer records array.](images/colr_layers_v0.png)
+
+**Figure 5.18 Version 0: Color glyphs are defined by slices of a layer records array.**
+
+The version 1 formats also define layer sets as slices in an array. In this case, however, the array is an array of offsets to paint tables, and is contained in a LayerV1List table. Each referenced paint table is the root of a sub-graph of paint tables that specifies a graphic composition to be used as a layer. Within a given slice, the first offset provides the content for the bottom layer, and each subsequent offset provides content that overlays the preceding content. Definition of the slices (the sub-sequence of offsets for each slice) are given in PaintColorLayers tables.
```
rsheeter, line 1259: I think this overemphasizes the use of slices of the layer list. You don't *have* to use slices of arrays at all, the core concept is really that it's a graph. I think that means we have to go directly to explaining the graph concept. Spitballing:

The version 1 format defines layer sets as a DAG of paints. The DAG is traversed and each Paint is processed when encountered. For all Paint formats except PaintComposite each subsequent record defines content that overlays the preceding content, just as with version 0. For PaintComposite the mode dictates how the compositing is done. Layers can be shared in three main ways in this model:

1. The Offset to a Paint may point to a Paint that is also referenced by other DAGs
1. PaintColrLayers takes a slice of the LayerV1List, enabling subsets of color glyphs to be shared
1. PaintColrGlyph allows reuse of the complete set of layers making up a color glyph

colr-gradients-spec.md
```diff
@@ -1246,6 +1246,44 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+
+![PaintColrLayers table used as the root of a color glyph definition.](images/colr_PaintColrLayers_as_root.png)
+
+**Figure 5.20 PaintColrLayers table used as the root of a color glyph definition.**
+
+A PaintColorLayers table can also be nested more deeply within the graph, providing a layer structure to define some component within a larger color glyph definition. (See 5.7.11.1.7.2 for more information.) The ability to nest a PaintColrLayers table within a graph creates the potential to introduce a cycle within the graph. This is not valid, however: graphs shall be acyclic.
+
+> **_TBD: Add reference to separate section discussing cycles (#156)._**
+
+Use of layers in a color glyph definition, and of the PaintColrLayers table specifically, is optional. While a PaintColrLayers can be used as the root of a color glyph, this is not required. For example, a complete color glyph definition might be comprised of a simple graph containing only glyph and basic fill tables, as illustrated in figure 5.21:
```
rsheeter, line 1279: You can have layers in the sense of z-ordered content without LayerV1List so instead of "Use of layers" I think you want "Use of the LayerV1List." Also, if you accept some of the suggestions above I think the sentence about it being optional is redundant because we've explained as much earlier.

colr-gradients-spec.md
PeterConstable: Well, you're proposed wording is mostly fine and mostly says the same thing, I think. However, for v1, while it is true that the vector of layers is used only in conjuction with PaintColrLayers, it is *not* true that these are used *only* for sharing between color glyphs. If you want a set of layers at all, you will use the LayerV1List + PaintColrLayers because the only other means of layering is via PaintComposite and that only gives two layers at a time.

PeterConstable: E.g., figure 5.20: no sharing involved there. Also figure 5.25. Figure 5.30 (later PR) has two PaintColrLayers: one for sharing, and one not.

colr-gradients-spec.md
PeterConstable: First half of that isn't about layers, so should be in this section. Second half is basically the intro already in place for 5.7.11.1.7.

colr-gradients-spec.md
PeterConstable: How do you create z-ordered content without using LayerV1List? The only other way would be using PaintComposite, and that's only two layers at a time. Nothing else does layering.

PeterConstable:
Rod & I chatted, and I think we're now on the same page. (Nuances in wording.) I've added a commit that hopefully addresses his concerns.

colr-gradients-spec.md
```diff
@@ -914,6 +914,98 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+Figure 5.18 illustrates layers using version 0 formats.
+
+![Version 0: Color glyphs are defined by slices of a layer records
+array.](images/colr_layers_v0.png)
+
+**Figure 5.18 Version 0: Color glyphs are defined by slices of a layer records
+array.**
+
+When using version 1 formats, layers are supported but are optional. For
+example, a simple glyph description need not use any layering, as illustrated in
```
rsheeter, line 940: To me this is still a layer so I dislike the phrasing that *layers* are optional. The optional thing is writing down the offset in the LayerV1List. Perhaps something more like "When using the version 1 format, recording layers in the LayerV1List is optional. For example, a simple glyph can be referenced directly from the BaseGlyphV1Record:"

colr-gradients-spec.md
```diff
@@ -914,6 +914,98 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+layers.](images/colr_color_glyph_without_layers.png)
+
+**Figure 5.21 Complete color glyph definition without use of layers.**
+
+The version 1 formats include two ways to add layers:
+
+* The PaintComposite table allows two sub-graphs to be composed together using
+different compositing or blending modes.
+
+* The PaintColrLayers table supports defining a sequence of several layers.
```
rsheeter, line 954: Shouldn't PaintColrGlyph be considered a third way, allows a complete color glyph graph to be reused as part of another color glyphs graph.

colr-gradients-spec.md
```diff
@@ -914,6 +914,98 @@ A PaintGlyph table on its own does not add content: if there is no child paint t
...
+providing a layering of content defined in
+sub-graphs.](images/colr_layers_v1.png)
+
+**Figure 5.19 Version 1: PaintColrLayers tables specify slices within the
+LayerV1List, providing a layering of content defined in sub-graphs.**
+
+NOTE: Paint table offsets in the LayerV1List table are only used in conjuction
+with PaintColrLayers tables. If a paint table does not need to be referenced via
+a PaintColrLayers table, its offset does not need to be included in the
+LayerV1List array.
```
rsheeter, line 987: Should we phrase this in specese? - Picturing something like: To be reusable by means of PaintColrLayers Paints shall be recorded in the LayerV1List. Paints not reused by PaintColrLayers should not be recorded in the LayerV1List.

rsheeter (approved):
Few nits, generally LGTM

colr-gradients-spec.md
PeterConstable: PaintColrGlyph doesn't itself increase the layers in the graph.

colr-gradients-spec.md
PeterConstable: In ISO-ese, a note shall not contain shall. Also, I don't think a font should be invalid if it has a paint offset in the LayerV1List but that is never used by any PaintColrLayers table.

colr-gradients-spec.md
PeterConstable: And this section isn't about re-use.

colr-gradients-spec.md
PeterConstable: Sure: there's always at least one layer. I've tweaked wording, including renaming the section "Layering".

drott (approved):
No objections from my side, but I'd leave it to @rsheeter to finalise the review here, as you and @rsheeter have been discussing this PR in more detail together.

colr-gradients-spec.md
rsheeter: > PaintColrGlyph doesn't itself increase the layers in the graph.

I don't quite follow. Think of it this way, it's the same as using PaintColrLayers to reference the complete set of DAG roots that are used by another glyph. 

For example, we have three color glyphs:

1. a BaseGlyph with it's Paint(s) listed in LayerV1List
1. a BaseGlyph that uses PaintColrGlyph to reference the first BaseGlyph
1. a BaseGlyph that uses PaintColrLayers to pull in all the layers of the first BaseGlyph

2 and 3 both pull in all the layers of 1.

colr-gradients-spec.md
```diff
@@ -889,7 +889,100 @@ To illustrate this, the example in figure 5.16 is extended in figure 5.17 so tha
...
+
+![Version 0: Color glyphs are defined by slices of a layer records
+array.](images/colr_layers_v0.png)
+
+**Figure 5.18 Version 0: Color glyphs are defined by slices of a layer records
+array.**
+
+When using version 1 formats, use of multiple layers is supported but is
+optional. For example, a simple glyph description need not use any layering, as
+illustrated in figure 5.21:
```
rsheeter, line 919: Curious what @drott and @anthrotype think because this still doesn't feel right to me. I would say the use of LayerV1List is optional but you always have layers. The simple graph only has 1 layer but it's still a layer.

colr-gradients-spec.md
PeterConstable: When you reference a DAG using PaintColrGlyph, you get as many layers as there are in the DAG instead of 1 layer. If the DAG has one layer, then you still have only 1 layer. PaintColrGlyph doesn't itself add layers. 

>it's the same as using PaintColrLayers to reference the complete set of DAG roots that are used by another glyph.

This is falling into the same mistake I made earlier, of assuming a DAG starts with a PaintColrLayers. If the DAG does start with PaintColrLayers, then it is PaintColrLayers that adds layers, not the PaintColrGlyph that referenced it. 

>a BaseGlyph with it's Paint(s) listed in LayerV1List

A BaseGlyphV1Record doesn't point into the LayerV1List; it points to a single paint. (You're thinking of the earlier design, before PaintColrLayers was introduced.) So, "Paint(s)" is incorrect. And there's no point in having the paint referenced by the BaseGlyphV1Record be included in the LayerV1List unless that same paint is somewhere else referenced by a PaintColrLayers.

colr-gradients-spec.md
PeterConstable: >use of multiple layers

One layer is not multiple layers.

This section is leading with the concepts of each graphic capability, and from there introduce the formats. In this case, the concept is _layering_, i.e. use of multiple layers, and the flow is to start with that concept and then explain how it's achieved. To start out saying "LayerV1List is optional" would change from that pattern by leading with the format.

colr-gradients-spec.md
rsheeter: > A BaseGlyphV1Record doesn't point into the LayerV1List; it points to a single paint.

This may be the core of how we're seeing this differently. In "a BaseGlyph with it's Paint(s) listed in LayerV1List" to me "it's Paints" means all the Paints reachable through the graph whose root the BaseGlyph points to. It sounds like you are taking it to mean just the one Paint the BaseGlyph points at (the root), not the entire graph?

rsheeter: Resolving pending new content per f2f. I now much better understand what was intended here, ty for explaining.

colr-gradients-spec.md
rsheeter: After our f2f chat I better understand where this is coming from, resolving.

colr-gradients-spec.md
```diff
@@ -889,7 +889,121 @@ To illustrate this, the example in figure 5.16 is extended in figure 5.17 so tha
...
+
+
+![Complete color glyph definition without use of
+layers.](images/colr_color_glyph_without_layers.png)
+
+**Figure 5.21 Complete color glyph definition without use of layers.**
+
+The version 1 formats define a color glyph as a graph of paint tables, and the
```
rsheeter, line 927: s/as a graph of paint/as a directed acyclic graph/; that detail is very important to some readers

colr-gradients-spec.md
```diff
@@ -889,7 +889,121 @@ To illustrate this, the example in figure 5.16 is extended in figure 5.17 so tha
...
+**Figure 5.20 PaintColrLayers table used as the root of a color glyph
+definition.**
+
+A PaintColorLayers table can also be nested more deeply within the graph,
+providing a layer structure to define some component within a larger color glyph
+definition. (See 5.7.11.1.7.2 for more information.) The ability to nest a
+PaintColrLayers table within a graph creates the potential to introduce a cycle
+within the graph. This is not valid, however: graphs shall be acyclic.
+
+> **_TBD: Add reference to separate section discussing cycles (#156)._**
```
rsheeter, line 1006: This should be 5.7.11.1.8 Color glyphs defined as a graph I think looking at master?

colr-gradients-spec.md
```diff
@@ -889,7 +889,121 @@ To illustrate this, the example in figure 5.16 is extended in figure 5.17 so tha
...
+concept of layering corresponds roughly to the number of distinct leaf nodes in
+the graph. The basic fill formats— PaintSolid, PaintLinearGradient and
+PaintRadialGradient—do not have child paint tables and so can only be leaf nodes
+in the graph. Some paint tables, such as the PaintGlyph table, have only a
+single child, so can be used within a layer but do not provide any means of
+adding additional layers. Increasing the number of layers requires paint tables
+that have two or more children, creating a fork in the graph.
+
+NOTE: In more precise terms, the number of visual layers represented by a valid
+graph is the number of distinct root to leaf paths in the graph.
```
rsheeter, line 937: If layering is conceptually the simple-alpha composition of things in a specific order then I think some PaintComposite modes escape this model, potentially meaning this isn't strictly true. If so I suggest this might need to say the number of visual layers is *at most* the number of leaves?

rsheeter (approved):
Liking this version much better, ty

colr-gradients-spec.md
PeterConstable: I have to add references for that in at least a couple of places, and also update some other numbering, so will take care of that in a later PR.

colr-gradients-spec.md
PeterConstable: Well, the same leaf node can potentially be used in multiple layers. So, it's not actually true that the number of visual layers is <= the number of leaf nodes.

But I hadn't considered if some PaintComposite modes complicate the statement. I've opened #165 to consider this further.

PR #140, opened 2020-12-05
Extend content in transformations section

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1226,12 +1226,31 @@ strip or a cone filled with a linear gradient.*
...
-in `src`. The transformation affects all nested drawing operations. It affects
-how nested solid paints and gradients are drawn, as well as how nested clip
-operations or nested COLR glyph reuse operations are performed.
+A PaintTransform table can be used within a color glyph description to apply an
+affine transformation matrix. Supported transformations using a matrix can be a 
+combination of: scale, skew, mirror, rotate, or translate. The transformation is
+applied to all nested paints in the child sub-graph, including a nested sub-graph
+referenced using a PaintColrGlyph table (see 5.7.11.1.7.3).
+
+NOTE: Affine transformations have the property that parallel lines remain parallel.
```
rsheeter, line 1235: I appreciate the intent but I'm skeptical that in isolation this is helpful; I think we should take the cowards way out and cite something that explains affines.

If we're allowed to cite wikipedia their section of affine is pretty good; https://en.wikipedia.org/wiki/Affine_transformation#/media/File:2D_affine_transformation_matrix.svg gives a nice summary of what you can achieve.

colr-gradients-spec.md
```diff
@@ -1226,12 +1226,31 @@ strip or a cone filled with a linear gradient.*
...
-transformation `transform` to the current drawing region.  The transformation is
-to be applied for subsequent nested paints, as defined by the paint referenced
-in `src`. The transformation affects all nested drawing operations. It affects
-how nested solid paints and gradients are drawn, as well as how nested clip
-operations or nested COLR glyph reuse operations are performed.
+A PaintTransform table can be used within a color glyph description to apply an
+affine transformation matrix. Supported transformations using a matrix can be a 
+combination of: scale, skew, mirror, rotate, or translate. The transformation is
+applied to all nested paints in the child sub-graph, including a nested sub-graph
+referenced using a PaintColrGlyph table (see 5.7.11.1.7.3).
```
rsheeter, line 1233: Nit: I think it's a stronger statement without the "including a..." bit. All means all, if you start talking about what is included that makes me think something else might not be.

colr-gradients-spec.md
```diff
@@ -1226,12 +1226,31 @@ strip or a cone filled with a linear gradient.*
...
+
+NOTE: Affine transformations have the property that parallel lines remain parallel.
+
+The effect of a PaintTransform table is illustrated in figure 5.21: a PaintTransform is used to specify a rotation, and both the glyph outline and gradient in the sub-graph are rotated.
+
+![A rotation transformation rotates the fill content defined by the child sub-graph.](images/colr_transform_glyph_gradient.png)
+
+**Figure 5.21 A rotation transformation rotates the fill content defined by the child sub-graph.**
```
rsheeter, line 1241: I didn't find the mirror-Y as obvious as 5.21's rotation. Suggestion: Rotate => Mirror => PaintGlyph => PaintLinearGradiant. Pretty hard to miss that the triangle turned over.

colr-gradients-spec.md
```diff
@@ -1226,12 +1226,31 @@ strip or a cone filled with a linear gradient.*
...
+
+![Combined effects of a transformation nested within the child sub-graph of another transformation.](images/colr_transform_glyph_transform_gradient.png)
+
+**Figure 5.22 Combined effects of a transformation nested within the child sub-graph of another transformation.**
+
+The affine transformation is specified in a PaintTransform table as matrix elements. See <span style="color:red">5.7.11.2.x</span> for format details.
+
+Whereas the PaintTransformed table supports several types of transforms, the PaintRotate and PaintSkew tables support specific transformations: rotation and skew. The significant difference of these paint formats is that rotations and skews are specified as angles, in counter-clockwise degrees. In many cases, specifying the rotation or skew as an angle can be more convenient. It can also have a signficant benefit in variable fonts if an angle of skew or rotation needs to vary: it is much easier to implement variation of angles when specified directly than when specified using matrix elements.
```
rsheeter, line 1251: Anything we could cite for "it is much easier to implement variation of angles when specified directly than when specified using matrix elements."? - if not no worries

colr-gradients-spec.md
```diff
@@ -1226,12 +1226,31 @@ strip or a cone filled with a linear gradient.*
...
+
+The PaintRotate and PaintSkew tables also provide another advantage for these specific types of transformation, compared to the PaintTransformed table. Rotations or skews implemented with a matrix are always relative to the origin. To get the desired result using only PaintTransformed tables, a rotation or skew matrix might need preceding and following translations to shift the origin when the rotation or skew is applied. In contrast, the PaintRotate and PaintSkew tables allow a center of rotation to be specified directly.
```
rsheeter, line 1253: This isn't right, you can compute a 2x3 that rotates (or skews or whatever) around an arbitrary origin. Any given instance of PaintRotate or Skew has an equivalent 2x3. The primary advantage is making it easy to vary angle. Saving bytes in some situations is a small secondary advantage.

rsheeter: Suggest just deleting L1253.

rsheeter (changes requested):
Approved subject to deletion of the section about PaintTransformed not being able to rotate or skew other than relative to origin. I ... think ... that means I should mark "Request changes."

colr-gradients-spec.md
PeterConstable: In general, I think there are lots of resources that describe matrix transformations that people can easily find. But finding the meaning of "affine" isn't as easy: for whatever reason, not many resources make a point of explaining that. I learned about matrix transformations in high school, but up through my math undergrad never came across the term "affine". I figured others might be in the same situation, but perhaps I'm accommodating too much to my special case.

colr-gradients-spec.md
PeterConstable: The reason I added that is because in the case of PaintColrGlyph there isn't a direct paint -> paint -> paint linkage. It might not be clear to some whether paint -> paint -> **_glyphID_** and glyphID -> paint -> paint comprise one graph or two. That should probably be stated somewhere, but perhaps this isn't the best place.

colr-gradients-spec.md
PeterConstable: The reason I did it as transform -> glyph -> transform -> gradient was so that there would be two visible elements affected by the first transform but only one visible element affected by the second.

colr-gradients-spec.md
PeterConstable: For rotation of a given angle, the matrix elements include sin and cos of the angle. It's easy to give a smooth and linear interpolation of the angle, but to do that for matrix elements would either require HOI or else many deltas over "intermediate" ranges. I'll take a stab at giving a rational without going into too much detail.

colr-gradients-spec.md
PeterConstable: @rsheeter You can use 2x3 matrices to rotate around an arbitrary point, but not using a single matrix. E.g., if you want to rotate using (say) the point (2, 3) as the center or rotation, using 2x3 matrices you would need to do this in three steps: translate (-2, -3), then rotate, then translate (2, 3). The rotation needs to be wrapped by distinct translations. The PaintRotate table allows this to be stated directly in a single table.

drott (commented):
Could you add some kind of source images for the illustrations, .eps, .pdf, .ai or similar? 

For Figure 5.22, may I suggest to add in the PaintTransform block to explain by how many degrees this transform rotates?

colr-gradients-spec.md
drott: The matrices A (translate), B (rotate), C (translate back) in your example can be combined into one equivalent affine combining the three, resulting in the same transform.

PeterConstable: OK, convinced myself of that by extending the 2x3 into 3x3 and composing those. (You can't directly compose 2x3s.)

colr-gradients-spec.md
rsheeter: Something terse to that effect strikes me as worth including, it's one of the least obvious things in here imho.

...just noticed the `<span>`; let's try not to mix html in if at all possible

PeterConstable: The <span> will be removed once I know what the clause ref will be.

colr-gradients-spec.md
rsheeter: I think the text you have about this enabling rotate, skew, etc will be helpful to someone who hasn't seen an affine transform. I suggest either delete the note or change it to be a reference: "See blah for more about affine transformations"

rsheeter: Maybe cite https://www.w3.org/TR/css-transforms-1/#mathematical-description and note how our fields map to the abcdef shown there

colr-gradients-spec.md
rsheeter: Don't *have* to change anything, I just found that unlike the other examples where you can see at a glance what's happening for this one I had to stop and look at it to spot what it was telling me

rsheeter:
If Dominik change is made and L1253 is deleted IMHO this could be merged

colr-gradients-spec.md
PeterConstable: We explain our matrix elements clearly, and cross-ref to CSS might only create confusion. I'll drop the note.

colr-gradients-spec.md
PeterConstable: Added a bit of text to call out the details.

colr-gradients-spec.md
PeterConstable: Removed here. Will consider mention elsewhere (cf. #151)

rsheeter (approved):

drott:
> Could you add some kind of source images for the illustrations, .eps, .pdf, .ai or similar?

@PeterConstable - what about vector source images for `images/colr_transform_glyph_gradient.png` and `images/colr_transform_glyph_transform_gradient.png` - would it be possible to add those?

PeterConstable:
>@PeterConstable - what about vector source images for images/colr_transform_glyph_gradient.png and images/colr_transform_glyph_transform_gradient.png - would it be possible to add those?

@drott Sorry, meant to reply: I assemble all of these images in a single Powerpoint file, which is binary. I'll add that after the last PR for Graphic components is merged.

PR #141, opened 2020-12-05
Extend content for Compositing and blending section

PeterConstable:
Mostly new content.


colr-gradients-spec.md
```diff
@@ -1235,13 +1235,36 @@ operations or nested COLR glyph reuse operations are performed.
...
+
+A PaintComposite table can be used to get other compositing or blending effects. The
+PaintComposite table combines content defined by two sub-graphs: a *source* sub-graph;
+and a destination, or *backdrop*, sub-graph. First, the paint operations
+for the backdrop sub-graph are executed, then the drawing operations for the source sub-graph are executed
+and combined with backdrop using a specified compositing or blending mode.
+Various composition and blending modes are supported. The available modes are given in the CompositionModes enumeration (see 5.7.11.2.x). The effect and processing rule of each mode are specified in [Compositing and Blending Level 1][1].
```
rsheeter, line 1245: Suggest deletion of "Various composition and blending modes are supported."; redudant with "can be used to get other compositing or blending effects."

colr-gradients-spec.md
```diff
@@ -1235,13 +1235,36 @@ operations or nested COLR glyph reuse operations are performed.
...
+
+**Figure 5.23 Two content elements combined using the Porter-Duff XOR mode.**
+
+Figure 5.24 shows the effect of a “lighten” blending mode, which has the effect that the R, G, and B color components for each pixel in the result is the greater of the R, G, and B values from corresponding pixels in the source and backdrop.
+
+![Two content elements combined using the lighten blending mode.](images/colr_blend_lighten.png)
+
+**Figure 5.24 Two content elements combined using the lighten blending mode.**
+
+For complete details on each of the Porter-Duff and blending modes, see the [Compositing and Blending Level 1][1] specification.
```
rsheeter, line 1261: Given that W3 Compositing has extensive examples I wonder if we need our own? - this feels to me like a case where just the citation might be enough, no need for L1247 through 1259.

Curious what @drott or @anthrotype think.

colr-gradients-spec.md
```diff
@@ -1235,13 +1235,36 @@ operations or nested COLR glyph reuse operations are performed.
...
+
+Figure 5.25 illustrates how the PaintComposite table is used in combination with content sub-graphs to implement an alternate compositing effect. The source sub-graph defines a green capital A; the backdrop sub-graph defines a black circle. The compositing mode used is “source out”, which has the effect that the source content punches out a hole in the backdrop. (For this mode, the fill color of the source is irrelevant; a black or yellow "A" would have the same effect.) A red rectangle is included as a lower layer to show that the backdrop has been punched out by the source, making that portion of the lower layer visible.
+
+![A color glyph using a PaintComposite table to punch out a shape from the fill of a circle.](images/colr_PaintCompositeGraph.png)
+
+**Figure 5.25 A color glyph using a PaintComposite table to punch out a shape from the fill of a circle.**
```
rsheeter, line 1267: Big fan of 5.25, I think that's a helpful example to have.

rsheeter (approved):
Approving; if we decide to tersen the modes discussion it's easy enough to circle back and delete some words later.

colr-gradients-spec.md
PeterConstable: My thought here was just to introduce. Some people will read the spec and want to get the basic idea of what the capabilities are, but won't be implementing and won't want to digest all of the W3C doc. It gives examples of all, but not a quick explanation of the basic idea of the compositing modes or of the blending modes.

vlevantovsky: The usual approach to referencing an external spec is to provide explicit reference without duplication of content. This way, when the referenced specification (and supporting implementations) evolve over time, the standard that contains the original reference doesn't need to be updated (e.g. mandatory normative references to base OFF standard ISO/IEC 14496-22 published in 2009 are still valid, and those references now cover all the latest layout features, variable fonts capabilities, etc.). 

Since the current W3C Compositing spec is a CR, it is possible that it may be updated prior to becoming the final W3C Recommendation, and including those latest updates by reference can be a benefit. However, this also presents certain risks, when the referenced specs and implementations are subjected to significant changes, and previously created content doesn't work anymore. (One recent example comes to mind, when CSS Font Module, albeit being a WD circa 2012, undergone significant changes in the way how layout and other font features were to be applied. Browsers were updated soon after the spec change was agreed upon, which was a breaking change for any then-existing early complex layout content.) Sometimes, the alternative approach is justified - when a particular way of feature implementation is needed to be preserved, the exact version reference _and_ duplication of content in the referencing specification can be justified.

PeterCon: >The usual approach to referencing an external spec is to provide explicit reference without duplication of content. 

There is no duplication of normative content. Two examples (of the many options) are given as general overview of the concepts. For normative content, explicit reference is given.

>Since the current W3C Compositing spec is a CR...

Valid concerns in general. In this case, though, I think it's very unlikely that the specification in the W3C doc for the compositing and blending modes we are using (which are explicitly enumerated) will ever undergo a breaking change: these compositing and blending modes are standard fare, documented in various places, and changes to the behaviours in this spec would only create confusion and bugs.

PR #153, opened 2020-12-09
Add content for re-use using PaintColrGlyph

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1326,6 +1315,17 @@ There are several ways in which elements of a color glyph description can be re-
...
+
+The glyph ID used may be that for a glyph outline, if there is an appropriate glyph outline that corresponds to this composition. But the glyph ID may also be greater than the last glyph ID used for outlines—that is, greater than or equal to the numGlyphs value in the &#39;maxp&#39; table (5.2.6). Such virtual base glyph IDs in the COLR table are only used within a PaintColrGlyph table, and are not related to glyph IDs used in any other tables.
+
+The example from 5.7.11.1.7.2 is modified to illustrate use of a PaintColrGlyph table. In the following figure, a PaintColrLayers table references a slice within the LayerV1List that defines the shared component. Now, however, this PaintColrLayers table is treated as the root of a color glyph definition for base glyph ID 63,163. The color glyph for the one o’clock emoji is defined with three layers, as before, but now the bottom layer uses a PaintColrGlyph table that references the color glyph definition for glyph ID 63,163.
```
drott, line 1322: > 63,163

I don't understand why the glyph ID in text and illustration has a comma in it?

drott (commented):
Generally looks good to me. I'd suggest to just have a single glyph id in the example, e.g. 63 - if I am not missing something.

colr-gradients-spec.md
PeterConstable: Sorry: N American number formatting (1000s grouping delimiter). Removing.

colr-gradients-spec.md
```diff
@@ -1330,6 +1319,45 @@ There are several ways in which elements of a color glyph description can be re-
...
 
+A third way to re-use components in color glyph definitions is to use a nested
+PaintColrGlyph table. The organizational structure involved is partially like
+that for re-use using nested PaintColrLayers tables: a layer set for the shared
+composition is defined using a PaintColrLayers table. However, this
+PaintColrLayers table is the root of a color glyph definition and is associated
+with a base glyph ID. Each time the shared composition is to be re-used, it is
+referenced in a color glyph definition by its base glyph ID using a
+PaintColrGlyph table. That reference graph for the shared composition is thereby
+incorporated into the graph of the color glyph defintion that uses is.
```
rsheeter, line 1330: It's not quite right that "composition *is* defined using a PaintColrLayers." For example, a compiler might decide what you want is to PaintColrGlyph's, one atop the other, and issue a PaintComposite. Sketch of revised paragraph:


A third way to re-use components in color glyph definitions is to use a nested
PaintColrGlyph table. This Paint pulls the complete DAG from another color glyph, referenced by the base glyph ID captured in the BaseGlyphV1Record for the other color glyph, into the current graph at its position. PaintColrLayers allows reuse of a subset of the layers from another color glyph. PaintColrGlyph pulls in an entire other color glyph.

rsheeter (approved):
LGTM subject to adjustment of the first paragraph to not state that PaintColrLayers is the *only* way to define composition

colr-gradients-spec.md
PeterConstable: Yeah, fair comment.

colr-gradients-spec.md
```diff
@@ -1035,6 +1035,46 @@ tables to point to different hours.**
...
+within the LayerV1List that defines the shared component. Now, however, this
+PaintColrLayers table is treated as the root of a color glyph definition for
+base glyph ID 63,163. The color glyph for the one o’clock emoji is defined with
+three layers, as before, but now the bottom layer uses a PaintColrGlyph table
+that references the color glyph definition for glyph ID 63163.
+
+![A PaintColrGlyph table is used to reference the shared clock face composition
+via a glyph ID.](images/colr_reuse_clock-face_PaintColrGlyph.png)
+
+**Figure 5.31 A PaintColrGlyph table is used to reference the shared clock face
```
rsheeter, line 1067: Optional nit: for the depiction of the LayerV1List table (which I suppose appears in several places) I think it might help to indicate which way it reads, something like an arrow pointing up or down, or a 0 at the top/bottom or a number beside each of the boxes with "Offset 0xDoesn'tMatter" or ... w/e really. Context: I initially read this picture backwards :) If you want to address but not in this PR that works.

rsheeter:
LGTM

colr-gradients-spec.md
PeterConstable: I did include indices in the first figure showing LayerV1List, figure 5.19, and then dropped that in following figures to reduce clutter and size (the figures start to get kind of wide, which can be a problem reading OT pages on a narrower screen).

Each figure with a layer list has ellipses between records. I can replace those arrows.

PR #155, opened 2020-12-09
Add content for re-use using PaintColrLayers

PeterConstable:


drott (approved):
LGTM, thanks for the additional illustrations, these are very helpful. For those, it'd be great to have vector source images in .ai, .pdf, or whatever suitable format. 

Since @rsheeter originally wrote the shape reuse parts, let's have his review on this PR as well.

rsheeter:
LGTM, merging since @drott  has also approved

PR #167, opened 2020-12-15
Add statement re ignoring PaintColrGlyph if no BaseGlyphV1Record is found

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1198,6 +1198,10 @@ to the numGlyphs value in the &#39;maxp&#39; table (5.2.6). Such virtual base
 glyph IDs in the COLR table are only used within a PaintColrGlyph table, and are
 not related to glyph IDs used in any other tables.
 
+When a PaintColrGlyph table is used, a BaseGlyphV1Record with the specified
+glyph ID is expected. If no BaseGlyphV1Record with that glyph ID is found, the
+PaintColrGlyph table shall be ignored.
```
rsheeter, line 1203: If an ignored glyph is meant to permit the graph to still render then we need to know if ignored glyphs are bounded or unbounded, how to composite them (basically what replaces it). That's ugly and really all it's doing is letting us produce bad results from bad fonts.

Let's just declare the entire graph being processed to be invalid in this case and leave it up to the implementation to decide what to do. It's broken to the same degree as something like an offset that points outside the COLR table (or outside the entire font).

Hm. If you follow the "it's as broken as an invalid offset" train of thought then we *could* let the spec remain silent, leaving it for implementations to decide what to do.

PeterConstable: OK, this is bigger than just PaintColrGlyph: also affects PaintColrLayers pointing to something invalid, or even a paint child forward offset: the missing sub-graph could be what keeps the visual bounded.

So, with the pseudo-code for checking for cycles, we currently say that, if a cycle is detected, then ignore the paint that introduces the cycle. But would that also be subject to the same issue, potentially leaving the visuals unbounded?

PeterConstable:
I'm going to drop this PR and come back to the right way to cover this after the discussion of visual boundedness has been add. This is covered in #168 .

colr-gradients-spec.md
rsheeter: > with the pseudo-code for checking for cycles, we currently say that, if a cycle is detected, then ignore the paint that introduces the cycle

That's not good, we should probably just deem the entire graph invalid

PR #169, opened 2020-12-16
Update header formats section

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1324,6 +1337,25 @@ V1 Header
...
 
+The BaseGlyphV1List and its subtables are only used in COLR version 1. The
+ItemVariationStore is only used in variable fonts and in conjunction with a
+BaseGlyphV1List and its subtables.
+
+A font that includes a BaseGlyphV1List can also include BaseGlyph and Layer
+records for compatibility with applications that only support COLR version 0.
+For applications that support COLR version 1, if a given base glyph is supported
+in the BaseGlyphV1List as well as in a BaseGlyph record, the data in the
+BaseGlyphV1List should be used.
```
drott, line 1348: Would it make sense to phrase this more generally? For applications that support COLRv1, they should prefer to use glyph definitions from COLRv1? 

Edit: This could be combined with the sentence below: "Applications should search for a base glyph ID first in the
BaseGlyphV1List, then if not found, search in the BaseGlyph records array, if
present." with a sub heading about how to prioritize glyph search / how to handle backwards compat?

Edit 2: What do you think about having two subsections, one about "Combining COLR v0 and COLR v1" - one saying: "Glyph lookup in a font with v0 and v1 glyphs"?  My idea is that subsections may help navigating what's to come in the text here.

colr-gradients-spec.md
```diff
@@ -1324,6 +1337,25 @@ V1 Header
 | Offset32 | layersV1Offset | Offset to LayerV1List table. |
 | Offset32 | itemVariationStoreOffset | Offset to ItemVariationStore (may be NULL). |
 
+The BaseGlyphV1List and its subtables are only used in COLR version 1. The
+ItemVariationStore is only used in variable fonts and in conjunction with a
+BaseGlyphV1List and its subtables.
```
drott, line 1342: It may be redundant or implied in conventions, but I suggest to describe that "not used" here means that the offsets are set to null, as opposed to just leaving those fields out of the table.

drott (commented):
The semantics look good to me, but some suggestions re structure.

colr-gradients-spec.md
PeterConstable: @drott See if the new commit resolves this.

colr-gradients-spec.md
```diff
@@ -1311,7 +1310,25 @@ given, unless otherwise indicated.
...
+The COLR table begins with a header. Two versions have been defined. Offsets in the header are from the start of the table.
+
+**5.7.11.2.1.1 COLR version 0**
+
+*COLR version 0:*
+
+| Type | Field name | Description |
+|-|-|-|
+| uint16 | version | Table version number—set to 1. |
+| uint16 | numBaseGlyphRecords | May be 0 in a version 1 table. |
```
rsheeter, line 1322: May be 0 in a v1 table strikes me as odd as the only documentation for for v0? Should it say "Number of Base Glyph Records." either as well as or instead of "May be 0..."

rsheeter: Similar concern for numLayerRecords

rsheeter (approved):

colr-gradients-spec.md
PeterConstable: Ugh! Those are copy/paste errors. Fixing...

PR #172, opened 2020-12-16
Update ColorIndex, ColorStop and ColorLine

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1361,42 +1361,73 @@ encoded here.
...
+*ColorLine table:*
 
 | Type | Name | Description |
 |-|-|-|
 | uint8 | extend | An Extend enum value. |
 | uint16 | numStops | Number of ColorStop records. |
 | ColorStop | colorStops[numStops] | |
 
+The colorStops array should be in increasing stopOffset order.
+
```
drott, line 1413: I haven't tested whether that's needed. But I am a bit hesitant to use "should" here, either we don't say anything, or we say shall and specify that color stop should be rejected if they are not sorted. @rsheeter, @clupo?

rsheeter: Since the offsets are variable we should either say more (they could vary into or out of order) or nothing.

rsheeter: Per F2F, amend to should be *processed* in sorted order.

anthrotype: FYI, the SVG spec for "Gradient Stops" says that they should be sorted

> Each gradient offset value is required to be equal to or greater than the previous gradient stop's offset value. If a given gradient stop's offset value is not equal to or greater than all previous offset values, then the offset value is adjusted to be equal to the largest of all previous offset values.

colr-gradients-spec.md
```diff
@@ -1463,42 +1463,77 @@ encoded here.
...
 |-|-|-|
 | uint16 | paletteIndex | Index for a CPAL palette entry. |
 | VarF2Dot14 | alpha | Variable alpha value. |
 
-Values for alpha outside [0.,1.] are reserved.
+A paletteIndex value of 0xFFFF is a special case, indicating that the text
+foreground color (as determined by the application) is to be used.
+
+The alpha.value is always set explicitly. Values for alpha outside the range
+[0., 1.] (inclusive) are reserved; values outside this range must be clipped. A
```
rsheeter, line 1483: I think this should be "values outside this range *shall* be clipped"

rsheeter (approved):

PR #176, opened 2020-12-16
Update PaintTransformed

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1618,17 +1629,27 @@ Affine2x3 record
...
 | VarFixed | dy | Translation in y direction. |
 
-The `Affine2x3` record is a 2x3 matrix for 2D affine transformations, so
-that for a transformation matrix _M_ and an extended starting vector _v = (x, y, 1)_
-the mapped vector _v′_ is calculated as _v′ = M * v_. That is:
+Matrix operations are of the form *v&#x2032; = Mv*, where *v* and *v&#x2032;*
+are vectors for positions in the design grid. The starting position vector *v*
+is an extended 3×1 column matrix with the value 1 as a third matrix element:
+*(x,y,1)*. The result vector *v&#x2032;* is a 2×1 column matrix
+*(x&#x2032;,y&#x2032;)*, calculated as follows:
```
rsheeter, line 1636: I'm a bit dubious the matrix bit helps anyone. Ideally we could just cite something "see blah for more about affine transforms."

We've clearly stated how our basis mapping works. We give the formula to compute a transformed point given an original point and a transform. Maybe we should just delete this paragraph, writing something terse like:

To apply an Affine2x3 to an input point (x, y) compute the output point (x', y') as:

x' = ...
y' = ...

colr-gradients-spec.md
```diff
@@ -1618,17 +1629,27 @@ Affine2x3 record
...
+described by the Affine2x3 record maps the basis vectors to _î&#x2032; = (xx,
+yx)_ 
+and _ĵ&#x2032; = (xy, yy)_, and translates the origin to _(dx, dy)_.
+
+When the transformed composition from the referenced paint table (and its
+sub-graph) is composed into the destination (represented by the parent of this
+table), the source design grid origin is aligned to the destination design grid
+origin. The transform can translate the source such that a pre-transform
+position (0,0) is moved elsewhere. The *post-transform* origin, (0,0), is
+aligned to the destination origin.
```
rsheeter, line 1652: Optional: if we shorten the matrix bit above it could make sense to just drop this. If all the reader wants is implementation then compute new coords per the (x', y') formula and move on. If the reader wants discussion of details ... maybe that's best as a citation.

rsheeter (approved):
Left a couple of suggestions but overall LGTM

colr-gradients-spec.md
PeterConstable: No, I'd like to leave this as it may not be obvious to something without much experience implementing 2D graphics. Things similar to this in the OT spec have created confusion before and have needed to be clarified. (E.g., <https://github.com/MicrosoftDocs/typography-issues/issues/191>.)

PR #177, opened 2020-12-17
Update PaintTranslate/Rotate/Skew

PeterConstable:
In the NOTEs for PaintRotate/PaintSkew, I dropped the mention of angles being more amenable to smooth rotation. That's covered in a note in 5.7.11.1.5 Transformations, and that note has more explanation as to why that assertion is true.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -1734,38 +1734,91 @@ and _ĵ′ = (xy, yy)_, and translates the origin to _(dx, dy)_.
 
 **5.7.11.2.5.8 Format 8: PaintTranslate**
 
+Format 8 is used to apply a translation to a sub-graph. The paint table that is
+the root of the sub-graph is linked as a child.
+
+See 5.7.11.1.5 for general information regarding transformations in a color
+glyph definition.
+
+*PaintTranslate table (format 8):*
+
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 8. |
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
+| Offset24 | paintOffset | Offset to a Paint subtable. |
 | VarFixed | dx | Translation in x direction. |
 | VarFixed | dy | Translation in y direction. |
 
+NOTE: Translation can also be represented using the PaintTransformed table. The
+PaintTranslate table provides a more compact representation when only
+translation is required.
+
+When the translated composition from the referenced paint table (and its
+sub-graph) is composed into the destination, the source design grid origin is
+aligned to the destination design grid origin. The translation will result in
+the pre-transform position (0,0) being moved elsewhere. The *post-transform*
+origin of the source, (0,0), is aligned to the destination origin.
 
 **5.7.11.2.5.9 Format 9: PaintRotate**
+
+Format 9 is used to apply a rotation to a sub-graph. The paint table that is the
+root of the sub-graph is linked as a child. The amount of rotation is expressed
+directly as an angle, and X and Y coordinates can be provided for the center of
+rotation.
+
+See 5.7.11.1.5 for general information regarding transformations in a color
+glyph definition.
+
+*PaintRotate table (format 9):*
 
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 9. |
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
+| Offset24 | paintOffset | Offset to a Paint subtable. |
 | VarFixed | angle | Rotation angle, in counter-clockwise degrees. |
 | VarFixed | centerX | x coordinate for the center of rotation. |
 | VarFixed | centerY | y coordinate for the center of rotation. |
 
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -1632,38 +1632,91 @@ and _ĵ′ = (xy, yy)_, and translates the origin to _(dx, dy)_.
...
 
+NOTE: Translation can also be represented using the PaintTransformed table. The
+PaintTranslate table provides a more compact representation when only
+translation is required.
+
+When the translated composition from the referenced paint table (and its
+sub-graph) is composed into the destination, the source design grid origin is
+aligned to the destination design grid origin. The translation will result in
+the pre-transform position (0,0) being moved elsewhere. The *post-transform*
+origin of the source, (0,0), is aligned to the destination origin.
```
rsheeter, line 1658: It would be nice to only say this once rather than in every Paint that produces an affine. Or, as noted in the PaintTransformed PR just don't say it because for the most part this is reasoned about as just bouncing points around.

colr-gradients-spec.md
```diff
@@ -1632,38 +1632,91 @@ and _ĵ′ = (xy, yy)_, and translates the origin to _(dx, dy)_.
...
+
 | Type | Field name | Description |
 |-|-|-|
 | uint8 | format | Set to 8. |
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
+| Offset24 | paintOffset | Offset to a Paint subtable. |
 | VarFixed | dx | Translation in x direction. |
 | VarFixed | dy | Translation in y direction. |
 
+NOTE: Translation can also be represented using the PaintTransformed table. The
```
rsheeter, line 1650: Suggest noting what the equivalent PaintTransformed would be.

colr-gradients-spec.md
```diff
@@ -1632,38 +1632,91 @@ and _ĵ′ = (xy, yy)_, and translates the origin to _(dx, dy)_.
...
-| Offset24 | paintOffset | Offset to a Paint subtable, from start of PaintRotate table. |
+| Offset24 | paintOffset | Offset to a Paint subtable. |
 | VarFixed | angle | Rotation angle, in counter-clockwise degrees. |
 | VarFixed | centerX | x coordinate for the center of rotation. |
 | VarFixed | centerY | y coordinate for the center of rotation. |
 
-*__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing an angle to be specified directly in degrees, which is more amenable to smooth variation.*
+NOTE: Rotation can also be represented using the PaintTransformed table. The
+important difference is in allowing an angle to be specified directly in
+degrees, rather than as changes to basis vectors.
```
rsheeter, line 1682: I think the point about allowing smooth rotation on an axis was important, suggest not removing.

rsheeter (changes requested):
Why remove mention of the value to variable fonts? - thats important for Rotate and Skew

colr-gradients-spec.md
PeterConstable: I want to say it at once since it might not be obvious to some readers that don't have experience in 2D graphic implementations.

colr-gradients-spec.md
PeterConstable: I'll do it for rotate/skew assuming rotation about the origin. (For other centers, it's overly complicated.)

colr-gradients-spec.md
PeterConstable: It was stated earlier in 5.7.11.1.5, but by spelling out equivalent matrix values it fits OK to do here as well.

colr-gradients-spec.md
```diff
@@ -1734,38 +1734,109 @@ and _ĵ′ = (xy, yy)_, and translates the origin to _(dx, dy)_.
...
-*__Note:__ Rotation can also be represented using the PaintTransformed table. The important difference is in allowing an angle to be specified directly in degrees, which is more amenable to smooth variation.*
+NOTE: Pure rotation can also be represented using the PaintTransformed table.
+For rotation about the origin, this could be done by setting matrix values as
+follows for angle &theta;: 
+
+* _xx_ = cos(&theta;)
+* _yx_ = sin(&theta;)
+* _xy_ = -sin(&theta;)
+* _yy_ = cos(&theta;)
+* _dx_ = _dy_ = 0
```
rsheeter, line 1789: Looking at https://en.wikipedia.org/wiki/Affine_transformation#/media/File:2D_affine_transformation_matrix.svg I think perhaps the - is on the x part of y basis (xy) where it should be on the y part of x basis (yx)?

* _yx_ = - sin(&theta;)
* _xy_ = sin(&theta;)

PeterConstable: Check the direction of the rotation: our format uses counter-clockwise degrees. That figure is using clockwise rotation. Sign of sine is opposite.

rsheeter (approved):

colr-gradients-spec.md
rsheeter: Ah you are right, forgot we said counter-clockwise when comparing.

PR #178, opened 2020-12-17
Update PaintComposite

PeterConstable:


colr-gradients-spec.md
```diff
@@ -1715,9 +1717,14 @@ CompositeMode enumeration
 | 25 | COMPOSITE_HSL_COLOR | See [color blend mode][27] |
 | 26 | COMPOSITE_HSL_LUMINOSITY | See [luminosity blend mode][28] |
 
-> **_[under construction—more to come]_**
+The supported modes are taken from the W3C [Compositing and Blending Level 1][1] specification. For details on each mode, including specifications of the required behaviors, see the W3C specification.
+
+The graphic compositions defined by the source and backdrop paint tables (and their respective sub-graphs) are rendered into bitmaps, and the source is blended into the backdrop using the specified composition mode.
```
rsheeter, line 1722: Curious why this needs to be said? - we've provided the compositing spec which spells out exactly how to compute the result.

rsheeter (approved):

colr-gradients-spec.md
PeterConstable: Actually, I don't think _that_ W3C spec spells that out, but that you'd have to go into the SVG or HTML specs to find that detail. It's probably not essential, but I don't think it hurts or detracts.

PR #182, opened 2020-12-21
Add alpha mask example to 'Compisiting and blending' section

PeterConstable:
Replaced the old "Alpha" section (move/edit) with an addition to 5.7.11.1.6 Compositing and blending. This gets rid of the last bit of the original "OFF changes" section.

With a figure inserted, following figures are renumbered. And since the same figure is used (to illustrate something different) in 5.7.11.1.9, that has a minor change to indicate the repeated figure.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -18,11 +18,6 @@ December 2019
     - [Solid](#solid)
   - [COLR Glyph](#colr-glyph)
   - [COLR Layers](#colr-layers)
-- [OFF Changes](#off-changes)
-  - [COLR table](#off-5711-colr--color-table)
-    - [Data structures](#colr-v1-data-structures)
-      - [Understanding COLR v1](#understanding-colr-v1)
-        - [Alpha](#alpha)
 - [Implementation](#implementation)
   - [Font Tooling](#font-tooling)
   - [Rendering](#rendering)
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -1195,13 +1168,19 @@ visible.
 
 **Figure 5.27 A color glyph using a PaintComposite table to punch out a shape from the fill of a circle.**
 
+NOTE: [Scalable Vector Graphics (SVG)][31] supports alpha channel masking using the &lt;mask&gt; element. The same effects can be implemented in COLR version 1 using a PaintComposite table by setting a pattern of alpha values in the source sub-graph and selecting the source in mode. This is illustrated in figure 5.28.
```
rsheeter, line 1171: Optional, I wonder if we should point out (in text or by updating the graphic) that a better (less bytes, same outcome) graph for 5.27 would have the A painted in either black or red, with the PaintSolid(green) dropped out of the graph entirely.

rsheeter (approved):
Nit: diff would have been nicer as two PRs, one for figure renumbering (which all but doesn't need review) and one for the mask example

colr-gradients-spec.md
PeterConstable: Changing...

Issue #168, opened 2020-12-15
Add discussion of well-formedness, validity


PeterConstable:
5.7.11.1.8 currently covers acyclicity, and indicates that paints that introduce a cycle should be ignored.

I discussed with Rod adding something mentioning that a PaintColrGlyph expects a corresponding BaseGlyphV1Record, and we had in mind to say that if there wasn't a matching record then the paint should be ignored.

But it occurred to Rod that dropping a sub-graph might leave the remaining graph visually unbounded. So, perhaps rather than ignoring a sub-graph an implementation might need to ignore the entire graph. Our sense was that the best thing might be to say that implementations MAY ignore the graph. (If we don't mention it at all, it might not occur to implementers that there's an issue they should be considering.)

So, after adding the section on boundedness, find the right part of the spec to discuss this.

rsheeter:
Per f2f, the color glyph is just invalid in several cases we say "ignored"

PeterConstable:
Added: PR #183

PR #183, opened 2020-12-21
Update 5.7.11.1.9 to cover well-formedness/validity

PeterConstable:
Coming back to #168 — this is the issue that, if a sub-graph is ignored (e.g., the glyphID in PaintColrGlyph is bad), then the remainder of the graph could be unbounded.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -1454,14 +1454,14 @@ graph, checking whether any paint table was already encountered within that
 path. The following pseudo-code algorithm can be used:
 
 ```
-    // called initially with the root paint and an empty set activePaints
-    function paintIsAcyclic(paint, activePaints)
-        if paint is in activePaints
+    // called initially with the root paint and an empty set pathPaints
+    function paintIsAcyclic(paint, pathPaints)
+        if paint is in pathPaints
           return false // cycle detected
-        add paint to activePaints
+        add paint to pathPaints
         for each childPaint referenced by paint as a child subtable
-          call paintIsAcyclic(childPaint, activePaints)
-        remove paint from activePaints
+          call paintIsAcyclic(childPaint, pathPaints)
+        remove paint from pathPaints
 ```
 
 For the graph to be valid, it shall also be visually bounded, as described in
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -1432,12 +1432,12 @@ Thus, the generalization that can be made regarding the relationship between the
 number of layers and the nature of the graph is that the number of distinct
 root-to-leaf paths will be greater than or equal to the number of layers.
 
-The following are required for the graph to be well-formed:
+The following are necessary for the graph to be well-formed and valid:
 
-* All subtable links shall be valid:
+* All subtable links shall satisfy the following criteria:
   * Forward offsets are within the COLR table bounds.
-  * If a PaintColrLayers table is present, a LayersV1List is present, and the
-referenced slice is within the length of the LayersV1List.
+  * If a PaintColrLayers table is present, then a LayersV1List is also present,
+and the referenced slice is within the length of the LayersV1List.
   * If a PaintColrGlyph table is present, there is a BaseGlyphV1Record for the
 referenced base glyph ID.
 * The graph shall be acyclic.
@@ -1468,7 +1468,7 @@ within that sequence. The following pseudo-code algorithm can be used:
 NOTE: Implementations can combine testing for cycles and other well-formedness
 requirements together with other processing for rendering the color glyph.
 
-For the graph to be valid, it shall be visually bounded, as described in
+For the graph to be valid, it shall also be visually bounded, as described in
 5.7.11.1.8.2.
 
 If the graph is not well formed, the entire color glyph definition should be
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -956,7 +956,10 @@ by the child sub-graph.
 
 **Figure 5.18 A PaintGlyph table defines a clip region for the composition defined by its child sub-graph.**
 
-A PaintGlyph table on its own does not add content: if there is no child paint table, then the PaintGlyph table represents no presentation content and shall be ignored.
+A PaintGlyph table on its own does not add content: if there is no child paint
+table, then the PaintGlyph table represents no presentation content and shall be
+ignored. This could result in the entire color glyph definition being invalid.
+See 5.7.11.1.9 for additional information.
 
 **5.7.11.1.4 Layering**
 
@@ -1309,6 +1312,10 @@ to the numGlyphs value in the &#39;maxp&#39; table (5.2.6). Such virtual base
 glyph IDs in the COLR table are only used within a PaintColrGlyph table, and are
 not related to glyph IDs used in any other tables.
 
+When a PaintColrGlyph table is used, a BaseGlyphV1Record with the specified
+glyph ID is expected. If no BaseGlyphV1Record with that glyph ID is found, the
+color glyph should be ignored. See 5.7.11.1.9 for additional information.
+
 The example from 5.7.11.1.7.2 is modified to illustrate use of a PaintColrGlyph
 table. In the following figure, a PaintColrLayers table references a slice
 within the LayerV1List that defines the shared component. Now, however, this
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -1309,6 +1312,10 @@ to the numGlyphs value in the &#39;maxp&#39; table (5.2.6). Such virtual base
 glyph IDs in the COLR table are only used within a PaintColrGlyph table, and are
 not related to glyph IDs used in any other tables.
 
+When a PaintColrGlyph table is used, a BaseGlyphV1Record with the specified
+glyph ID is expected. If no BaseGlyphV1Record with that glyph ID is found, the
+color glyph should be ignored. See 5.7.11.1.9 for additional information.
```
rsheeter, line 1317: s/should be ignored/is invalid/

colr-gradients-spec.md
```diff
@@ -956,7 +956,10 @@ by the child sub-graph.
 
 **Figure 5.18 A PaintGlyph table defines a clip region for the composition defined by its child sub-graph.**
 
-A PaintGlyph table on its own does not add content: if there is no child paint table, then the PaintGlyph table represents no presentation content and shall be ignored.
+A PaintGlyph table on its own does not add content: if there is no child paint
+table, then the PaintGlyph table represents no presentation content and shall be
+ignored. This could result in the entire color glyph definition being invalid.
```
rsheeter, line 961: I think we should directly say if there is no child paint then it is invalid, or alternately (preferably I think ... write it once) remove the text above and just point the reader to the section on validity

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,56 @@ Thus, the generalization that can be made regarding the relationship between the
...
-The following pseudo-code algorithm can be used:
+The following are required for the graph to be well-formed:
+
+* All subtable links shall be valid:
+  * Forward offsets are within the COLR table bounds.
+  * If a PaintColrLayers table is present, a LayersV1List is present, and the
+referenced slice is within the length of the LayersV1List.
+  * If a PaintColrGlyph table is present, there is a BaseGlyphV1Record for the
+referenced base glyph ID.
+* The graph shall be finite and acyclic (without cycles).
```
rsheeter, line 1443: (without cycles) is redundant

rsheeter: suggest just "The graph shall be acyclic"

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,56 @@ Thus, the generalization that can be made regarding the relationship between the
...
+* All leaf nodes shall be one of PaintSolid, PaintLinearGradient or
+PaintRadialGradient.
```
rsheeter, line 1445: Consider deletion or moving to a note; since we've required all the pointers to be valid it must be true that in a valid graph leaf nodes have no pointers. That is, if you implemented the checks above then this must be true.

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,56 @@ Thus, the generalization that can be made regarding the relationship between the
...
     // called initially with the root paint and an empty set activePaints
     function paintIsAcyclic(paint, activePaints)
         if paint is in activePaints
-          return failure (paint shall be ignored)
+          return false // cycle detected
         add paint to activePaints
         for each childPaint referenced by paint as a child subtable
           call paintIsAcyclic(childPaint, activePaints)
         remove paint from activePaints
 ```
```
rsheeter, line 1465: In reading this again I think we should have a note that this can and likely would be tracked as you process rendering. Assume the graph is valid, mosey on down it, and if you spot a cycle declare failure and bail.

colr-gradients-spec.md
rsheeter: I'd particularly like to not have "ignoring" be something we talk about at all.

colr-gradients-spec.md
```diff
@@ -956,7 +956,9 @@ by the child sub-graph.
 
 **Figure 5.18 A PaintGlyph table defines a clip region for the composition defined by its child sub-graph.**
 
-A PaintGlyph table on its own does not add content: if there is no child paint table, then the PaintGlyph table represents no presentation content and shall be ignored.
+A PaintGlyph table on its own does not add content: if there is no child paint
+table, then the graph is not well formed. See 5.7.11.1.9 for details regarding
+well-formedness and validity of the graph.
```
drott, line 961: Nit: My diff shows me an extra newline here. Would this be a Windows line endings artefact?

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,60 @@ Thus, the generalization that can be made regarding the relationship between the
 number of layers and the nature of the graph is that the number of distinct
 root-to-leaf paths will be greater than or equal to the number of layers.
 
-There is one constraint imposed on the complexity of the graph: the graph shall
-be acyclic—without cycles. That is, a paint table shall not have any child or
-descendent paint table that is also its parent or ancestor within the graph.
+The following are required for the graph to be well-formed:
```
drott, line 1435: I find it confusing from here and below that we use "well-formed" and valid in a slightly mixed way. Perhaps we can distinguish well-formedness as the static requirement that paint offsets need to be present and not out of bounds, and valid meaning that at runtime, no cycles are detected? Or distinguish them in other ways, but perhaps introduce and defined the terms. 

Perhaps it makes sense to say that we need well-formedness - i.e. no missing or out of bounds offset references, but for validity, we need cycle free

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,60 @@ Thus, the generalization that can be made regarding the relationship between the
...
+
+* All subtable links shall be valid:
```
drott, line 1437: As a suggestion to stay consistent in the terminology of offsets and paint tables that we use elsewhere and avoid the word _valid_ here.

```suggestion
* All references to sub-paint tables must meet the following criteria:
```

colr-gradients-spec.md
PeterConstable: No: I broke the line at 80 columns, as you've asked me to do.

colr-gradients-spec.md
PeterConstable: Well formed and valid are distinct notions. Well-formedness has to do with syntactic requirements. Validity has to do with semantic requirements. That is how I am using the terms.

Do you think of acyclicity as syntactic or semantic? (I was thinking of it as syntactic.)

colr-gradients-spec.md
PeterConstable: In this case, I think this is a semantic issue.

PeterConstable: Changing...

colr-gradients-spec.md
PeterConstable: I'll tweak wording...

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,60 @@ Thus, the generalization that can be made regarding the relationship between the
...
-table, checking whether any child was already encountered within that sequence.
-The following pseudo-code algorithm can be used:
+For the graph to be acyclic, no paint table shall have any child or descendent
+paint table that is also its parent or ancestor within the graph. In particular,
+because the PaintColrLayers and PaintColrGlyph tables use indirect child
+references rather than forward offsets, they provide a possibility for
+introducing cycles. Applications can test for cycles by tracking paint tables
+used in a sequence within the graph and then, when processing a PaintColrLayers
+or PaintColrGlyph table, checking whether any child was already encountered
+within that sequence. The following pseudo-code algorithm can be used:
```
rsheeter, line 1455: Suggest "path" over sequence within the graph and removal of specific reference to PaintColrLayers/PaintColrGlyph in "Applications can test..." because you have to check all formats, not just those. I suppose one could *not* check if you haven't seen one of those in a position other than root (a graph of only forward offsets can't cycle) but that is probably too much detail.

Perhaps replace "Applications can test..." with something like "Applications should track the active path while traversing the graph and fail if a Paint that exists in the active path is encountered a second time."

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,60 @@ Thus, the generalization that can be made regarding the relationship between the
...
+
+If the graph is not well formed, the entire color glyph definition should be
+considered invalid. In particular, applications should not simply ignore a
+sub-graph when an error in a paint table is detected since that could result in
+a visually unbounded color glyph definition. For instance, if a PaintColrGlyph
+table is found to introduce a cycle or there isn't a corresponding
+BaseGlyphV1Record, the remainder of the graph if that PaintColrGlyph table and
+its sub-graph are ignored could be unbounded. For this reason, if the graph is
+not well formed, the entire color glyph definition should be ignored. If the
+base glyph ID has an outline, that may be rendered instead.
```
rsheeter, line 1482: I think L1474..1482 could be terser. For example:

If the graph is not well formed, the entire color glyph definition should be
considered invalid. If the base glyph ID has an outline, that may be rendered instead.

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,60 @@ Thus, the generalization that can be made regarding the relationship between the
...
+
+Similarly, if an application encounters a paint table with an unrecognized
+format (which could be introduced in a later minor version update of the COLR
+table), the remainder of the graph could be visually unbounded, therefore the
+color glyph should be ignored.
```
rsheeter: This one is interesting. We could be denying ourselves some potentially desirable fallback strategies around degraded but still useful rendering if a format is unsupported. For example, imagine we add mesh gradients and you want to define a font that draws something useful on a mesh-incapable environment. Maybe that deserves more thought separate of this PR.

PeterConstable: We've moved the text in the direction of "should ignore the color glyph". But when we discussed this earlier, we talked about wording along the lines of "be aware the remainder may be unbounded... MAY ignore".

rsheeter: Ah and this is exactly why we had "ignored." OK, so we *should* say that, but in a clearly defined way.

rsheeter: Painting *nothing at all* is always bounded (same rationale as COMPOSITE_CLEAR); I think that's what we want for: cycles, bad PaintColrGlyph's, and unknown paint formats

PeterConstable: I've rev'd wording again to move in the direction we had discussed earlier: "be aware... may ignore..."

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,51 @@ Thus, the generalization that can be made regarding the relationship between the
...
+
+For the graph to be valid, it shall also be visually bounded, as described in
+5.7.11.1.8.2.
+
+If a sub-graph is ignored, the remainder of the graph could be visually
+unbounded and, therefore, invalid. This should be considered when devising a
+fallback strategy if an errors are detected while parsing the graph, or an
+unrecognized paint format is encountered. Applications may ignore the color
+glyph entirely in such cases. If the base glyph ID has an outline, that may be
+rendered as a non-color glyph instead.
```
rsheeter, line 1478: Apologies again for reversing myself but let's be more explicit about new paints in particular. Specifically, if a sub-graph is ignored or an unknown paint format is encountered nothing should be painted and the result shoudl be considered bounded.

colr-gradients-spec.md
PeterConstable: Thanks for the chat, which clarified for me what you were saying.

colr-gradients-spec.md
```diff
@@ -1425,31 +1432,61 @@ Thus, the generalization that can be made regarding the relationship between the
...
+For the graph to be acyclic, no paint table shall have any child or descendent
+paint table that is also its parent or ancestor within the graph. In particular,
+because the PaintColrLayers and PaintColrGlyph tables use indirect child
+references rather than forward offsets, they provide a possibility for
+introducing cycles. Applications should track paint tables within a path in the
+graph, checking whether any paint table was already encountered within that
+path. The following pseudo-code algorithm can be used:
 
 ```
     // called initially with the root paint and an empty set activePaints
```
rsheeter, line 1457: nit: suggest renaming activePaints to path for consistency with above

rsheeter (approved):

colr-gradients-spec.md
PeterConstable: Will use "pathPaints"

PR #185, opened 2020-12-22
Add draft for rendering algorithm

PeterConstable:
Draft of rendering algorithm -- let's discuss

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -2107,52 +2107,49 @@ NOTE: Checks for well-formedness and validity, as described in 5.7.11.1.9, are n
  
 ```
 // render a paint table and its sub-graph
-function renderPaint(paint, surface)
+function renderPaint(paint)
 
     if format 1: // PaintColrLayers
         for each referenced child paint table, in bottom-up z-order (see 5.7.11.1.4, 5.7.11.2.5.1):
-            create a new surface
-            call renderPaint() passing the child paint table and the new surface
-            compose the new surface with the existig surface using simple alpha blending
+            call renderPaint() passing the child paint table
+            compose the returned graphic onto the surface using simple alpha blending
 
     if format 2: // PaintSolid
         paint the specified color onto the surface
 
     if format 3: // PaintLinearGradient
     or if format 4: // PaintRadialGradient
-        paint gradient onto the surface following the gradient algorithm
+        paint the gradient onto the surface following the gradient algorithm
 
     if format 5: // PaintGlyph
         use the referenced glyph outline to set a clip region
-        call renderPaint() passing the child paint table and the existing surface
+        call renderPaint() passing the child paint table
         restore the previous clip region
 
     if format 6: // PaintColrGlyph
-        call renderPaint() passing the paint table referenced by the base glyph ID and the existing surface
+        call renderPaint() passing the paint table referenced by the base glyph ID
 
     if format 7: // PaintTransformed
     or if format 8: // PaintTranslate
     or if format 9: // PaintRotate
     or if format 10: // PaintSkew
         apply the specified transform
-        call renderPaint() passing the child paint table and the existing surface
+        call renderPaint() passing the child paint table
         restore the previous transform state
 
     if format 11: // PaintComposite
 
         // render backdrop sub-graph
-        create a new surface for the backdrop
-        call renderPaint() passing the backdrop child paint table and the backdrop surface
+        call renderPaint() passing the backdrop child paint table and save the result
 
         // render source sub-graph
-        create a new surface for the source
-        call renderPaint() passing the source child paint table and the source surface
+        call renderPaint() passing the source child paint table and save the result
 
         // compose source and backdrop
-        composing the source and backdrop surfaces using the specified composite mode
+        compose the source and backdrop using the specified composite mode
 
         // compose final result
-        compose the result of the above composition onto the existing surface that was passed in using simple alpha blending
+        compose the result of the above composition onto the surface using simple alpha blending
 ```
 
 **5.7.11.4 COLR table and OpenType Font Variations**
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -2125,6 +2125,7 @@ function renderPaint(paint, surface)
     if format 5: // PaintGlyph
         use the referenced glyph outline to set a clip region
         call renderPaint() passing the child paint table and the existing surface
+        restore the previous clip region
 
     if format 6: // PaintColrGlyph
         call renderPaint() passing the paint table referenced by the base glyph ID and the existing surface
@@ -2135,6 +2136,7 @@ function renderPaint(paint, surface)
     or if format 10: // PaintSkew
         apply the specified transform
         call renderPaint() passing the child paint table and the existing surface
+        restore the previous transform state
 
     if format 11: // PaintComposite
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -2096,83 +2096,61 @@ source and backdrop as follows:
 
 **5.7.11.3 COLR version 1 rendering algorithm**
 
-The following algorithm is used to render to render color glyphs defined using version 1 formats. The following assumes that the graph for the color glyph definition has been confirmed to be well formed and valid (see 5.7.11.1.9).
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -2094,7 +2094,86 @@ source and backdrop as follows:
 * Bounded *if and only if* both the source *and* backdrop are bounded:
   * All other modes
 
-**5.7.11.3 Rendering algorithm**
+**5.7.11.3 COLR version 1 rendering algorithm**
+
+The following algorithm is used to render to render color glyphs defined using version 1 formats. The following assumes that the graph for the color glyph definition has been confirmed to be well formed and valid (see 5.7.11.1.9).
```
rsheeter, line 2099: Let's be a little cautious about "is used"; it's not necessarily the only possible approach. How about we say something like "could be used."

Instead of saying we assume pre-confirmation of validity let's just say we elide validity checks for brevity because it's entirely valid to validate as you go.

colr-gradients-spec.md
```diff
@@ -2094,7 +2094,64 @@ source and backdrop as follows:
...
+    if format 2: // PaintSolid
+        paint the specified color onto the surface
+
+    if format 3: // PaintLinearGradient
+    or if format 4: // PaintRadialGradient
+        paint gradient onto the surface following the gradient algorithm
+
+    if format 5: // PaintGlyph
+        use the referenced glyph outline to set a clip region
+        call renderPaint() passing the child paint table and the existing surface
```
drott, line 2127: In this form, releasing the clip region is missing.

colr-gradients-spec.md
```diff
@@ -2094,7 +2094,64 @@ source and backdrop as follows:
...
+
+    if format 6: // PaintColrGlyph
+        call renderPaint() passing the paint table referenced by the base glyph ID and the existing surface
+
+    if format 7: // PaintTransformed
+    or if format 8: // PaintTranslate
+    or if format 9: // PaintRotate
+    or if format 10: // PaintSkew
+        apply the specified transform
+        call renderPaint() passing the child paint table and the existing surface
```
drott, line 2137: Restoring state to pre-transformed is missing here as the last step.

colr-gradients-spec.md
```diff
@@ -2094,7 +2094,64 @@ source and backdrop as follows:
...
+        // render backdrop sub-graph
+        create a new surface for the backdrop
+        call renderPaint() passing the backdrop child paint table and the backdrop surface
+
+        // render source sub-graph
+        create a new surface for the source
+        call renderPaint() passing the source child paint table and the source surface
+
+        // compose source and backdrop
+        composing the source and backdrop surfaces using the specified composite mode
```
drott, line 2150: These steps are very implementation specific and to some extend up to the 2D library for which operations it needs extra backing surfaces and for which it can do without. You can see from the way I wrote the algorithm earlier that with leaning towards how it's done in Skia, the API ergonomics are quite different to what is described here.

drott (commented):
With comments below, and as discussed in chat, I tend to agree with @rsheeter that our best option might be not to add this to the spec. This is an optional, rather coarse guideline for an option to implement this, in the way the algorithm was written before and is written here. What is your opinion, @PeterConstable - do you prefer to add this or not?

colr-gradients-spec.md
PeterConstable: The intent is to describe a logical behaviour that produces certain results, not reflect any particular 2D library.

colr-gradients-spec.md
PeterConstable: We need to include enough to specify what the expected behaviour is. Otherwise, implementations _will_ inevitably differ.

PeterConstable:
We need to include enough to specify what the expected behaviour is. Otherwise, implementations will inevitably differ.

PeterConstable:
"The following algorithm can be used to render color glyphs defined using version 1 formats. Applications are not required to implement rendering using this algorithm, but shall produce equivalent results."

colr-gradients-spec.md
drott: Sure, but "create a new surface" is one quite specific implementation strategy already / one particular implementation approach. When describing the logical behaviour, we just need wording to describe how to achieve the the intended outcome, not mandating how many backing layers or intermediate surfaces are created. 

You've choosen to add the surface as an argument to renderPaint() but the 2D libraries I have worked with (not only Skia) would not do it that way but instead they distinguisg separate concepts and avoid passing backing surfaces around: backing surface, canvas/drawing API on top of backing surface, state operations for manipulating compositing and transform state and managing layers. Logically, I agree there is not a huge difference between algorithm that has explicit surface arguments, but for the reader who wants to map it to their API concepts, perhaps an approach without passing those it's more straight-forward? 

Could we describe more in such a way? I.e. remove the surface argument to renderPaint() and talk about "set transform state, reset transform state " (as we do above),  "createLayer with composite mode" / renderChild / "mergeLayer and reset composite mode"?

Also, the final step you added here `// compose final result` is often be done in-place in the merge step - not sure it's needed in a generic algorithmic description. Perhaps it could be combined with the previous step: "mergeLayer while respecting alpha".

PeterConstable: The text explicitly says, "Applications are not required to implement rendering using this algorithm". Again, the intent is to describe enough to make clear what the expected behaviour/result is. Is this so different from the way that Skia or other libraries work that a reader will end up misunderstanding the intent?

>Also, the final step you added here // compose final result is often be done in-place in the merge step

Since the intent is explicitly not to tell how to implement, I don't see that this matters. If we state simply "mergeLayer", that is vague, and implementers will end up interpreting it in different ways.

drott: > Is this so different from the way that Skia or other libraries work that a reader will end up misunderstanding the intent?

That's the question, perhaps not. Of course we have freedom to describe it in any way we want if we say "applications do not have to follow this algorithm." But then why do we add this section? To give an easy to follow explanation of what would be a hopefully common way to implement it in common libraries - that's where my preference comes from. 

> Since the intent is explicitly not to tell how to implement, I don't see that this matters. If we state simply "mergeLayer", that is vague, and implementers will end up interpreting it in different ways.

Ok, with me.

I'll defer to @rsheeter for final approval as he also started the review. In the interest of not blocking anything when I am ooo, I don't object to merging this, despite not being quite content with how it's phrased atm.

PeterConstable: I'm making changes to remove mentions of new surfaces. I think it can still make the expected result clear, and hopefully helps in relation to your concerns.

colr-gradients-spec.md
```diff
@@ -2094,7 +2094,63 @@ source and backdrop as follows:
...
+**5.7.11.3 COLR version 1 rendering algorithm**
+
+The various graphic concepts represented by COLR version 1 formats were individually described in 5.7.11.1, and the various formats were described in 5.7.11.2. Together, these provide most of the necessary details regarding how a color glyph is rendered. The following provides a comprehensive description of the rendering process, considering the graph as a whole. 
+
+The following algorithm can be used to render color glyphs defined using version 1 formats. Applications are not required to implement rendering using this algorithm, but shall produce equivalent results.
+
+NOTE: Checks for well-formedness and validity, as described in 5.7.11.1.9, are not repeated here. Actual implementations can integrate such checks with rendering processing.
+
+1. Start with an initial drawing surface. As mentioned in 5.7.11.1.8.2, the bounding box of the base glyph can be used to determine the size.
+1. Traverse the graph of a color glyph definition starting with the root paint table referenced by a BaseGlyphV1Record. For each paint table, the following pseudo-code function is called recursively.
```
rsheeter, line 2106: I think you could delete "For each paint table, the following pseudo-code function is called recursively."; the recursion is directly stated in the pseudo-code

rsheeter:
This one is fascinating. New phrasing is much better (ty). However, I'm not entirely convinced anyone would implement correctly because of this where they otherwise would have done something wrong.

To keep us moving forward I suggest we merge this now, re-read the final result once Peter finishes his current wave of edits, and if it feels out of place or unhelpful we can then delete it.

rsheeter (approved):

colr-gradients-spec.md
drott: > I'm making changes to remove mentions of new surfaces.

Thanks for making these changes, looks better to me.

PR #186, opened 2020-12-22
Add section on variations

PeterConstable:
Last sub-clause to be added to 5.7.11 COLR

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -2099,9 +2099,10 @@ source and backdrop as follows:
 **5.7.11.4 COLR table and OFF Font Variations**
 
 The COLR table can be used in variable fonts. For color glyphs defined using
-version 0 formats, the glyph outlines can be variable. For color glyphs defined
-using version 1 formats, items that can be variable include the glyph outlines
-plus other aspects of the color glyph definition:
+version 0 formats, the glyph outlines can be variable, but no other aspect of
+the color glyph is variable. For color glyphs defined using version 1 formats,
+items that can be variable include the glyph outlines plus other aspects of the
+color glyph definition:
 
 * Alpha values
 * Color stop offsets in gradient color lines
@@ -2121,12 +2122,12 @@ data within the Item Variation Store. The record formats used include:
 * VarF2Dot14
 * VarFixed
 
-These are described in 7.2.3.1. For example, in the ColorStop record, the stop
-offset is represented using a VarF2Dot14 record, which has the following format:
+These are described in 7.2.3.1. They all follow a simple pattern: For a field
+type *SomeType* (hypothetical), the record format is as follows:
 
 | Type | Name | Description |
 |-|-|-|
-| F2Dot14 | value | |
+| *SomeType* | value | |
 | uint16 | varOuterIndex | |
 | uint16 | varInnerIndex | |
 
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -2096,13 +2096,51 @@ source and backdrop as follows:
 
 **5.7.11.3 Rendering algorithm**
 
-**5.7.11.4 COLR table and OpenType Font Variations**
+**5.7.11.4 COLR table and OFF Font Variations**
 
+The COLR table can be used in variable fonts. For color glyphs defined using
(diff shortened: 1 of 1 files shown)


drott (approved):
LGTM, but it'd be good if @anthrotype could have a look at this section as well. He's ooo atm, so perhaps he can take a look later and we merge in the meantime. @rsheeter, wdyt?

PeterConstable:
I have one small reservation with this draft, which is the repetition of the table defining the VarF2Dot14 record. 

I've included it because COLR v1 is introducing a new way of associating variable items with their deltas, not previously used, and even people familiar with OT formats and variable fonts could be confused. I exchanged mail with one person very familiar with current variable formats who was confused about how variation worked for COLR v1, and spelling out one of the Var* record formats made it clear. 

But in general, it's not good (wrt spec maintenance) to repeat a format definition in another part of the spec.

colr-gradients-spec.md
```diff
@@ -2096,13 +2096,51 @@ source and backdrop as follows:
 
 **5.7.11.3 Rendering algorithm**
 
-**5.7.11.4 COLR table and OpenType Font Variations**
+**5.7.11.4 COLR table and OFF Font Variations**
 
+The COLR table can be used in variable fonts. For color glyphs defined using
+version 0 formats, the glyph outlines can be variable. For color glyphs defined
```
rsheeter, line 2102: Suggest spelling out "glyph outlines can be variable but COLR table content is not" for v0 to make it more clear that the change is that v1 lets all the COLR content vary.

rsheeter:
+1 to @drott suggestion to merge, worst case we can make minor edits as it winds its way through ISO.

> I have one small reservation with this draft, which is the repetition of the table defining the VarF2Dot14 record.

Would making the example not repeat an actual definition help? - e.g. These formats follow a simple pattern. For a field of type SomeType a record VarSomeType is defined, having the following format:

| Type | Name | Description |
|-|-|-|
| SomeType | value | The initial value, prior to application of variation data (if any exists). |
| uint16 | varOuterIndex | Outer index into Item Variation Store or 0xFFFF if no variation data exists |
| uint16 | varInnerIndex | Inner index into Item Variation Store or 0xFFFF if no variation data exists |

rsheeter (approved):

PeterConstable:
>Would making the example not repeat an actual definition help? 

Yeah, that helps.

colr-gradients-spec.md
```diff
@@ -2096,13 +2096,52 @@ source and backdrop as follows:
...
+
+The value field of this record provides the default value for the stop offset.
+The remaining fields provide index values for a particular ItemVariationData
+subtable and DeltaSet record—the two-level organizational hierarchy used within
+the Item Variation Store.
+
+The index fields of the VarFWord, VarUFWord, VarF2Dot14 and VarFixed records
+shall always be set with specific values. The indices are base 0, therefore
+0x0000 cannot be used as an ignorable default. To indicate that an item has no
+variation data, the index fields shall be set to 0xFFFF/0xFFFF. (See 7.2.3.2.)
```
anthrotype, line 2142: > 0x0000 cannot be used as an ignorable default
Oh.. In fonttools colorLib I was actually doing exactly that: i.e. defaulting varIdx to 0 when not explicitly set. I thought that the precise value could simply be ignored when the COLR table has no ItemVariationStore.

https://github.com/fonttools/fonttools/blob/d0ea03c23b6eb81f80f0f8c6d85ae2119a759bf3/Lib/fontTools/ttLib/tables/otTables.py#L1258

So IIUC I should now change the default value of varIdx to 0xFFFF to mean variations.

I wonder if the "shall" could be relaxed a bit. We could say that the varIdx value only has an effect if the COLR table contains an ItemVariationStore, otherwise it can be ignored (and a compiler can use whatever is most convenient). I believe zeros compress away better than 0xFFFF (but haven't tested that yet empirically).

rsheeter: +1 to explicitly stating if there is no varstore then ignore the index fields and take the value

PeterConstable: @anthrotype Probably best to open a separate issue for this.

PeterConstable: I've opened #200

PR #192, opened 2020-12-23
Add PPTX with images

PeterConstable:
The PPTX is a source for several images.

The .md file extracts the content for the proposed update of OFF 5.7.11 into a separate file and adds a TOC, for convenience of reviewers.


rsheeter:
Best to delete (maybe add a link) the OFF content from colr-gradients-spec.md; don't want two copies

PeterConstable:
No, this is only a snapshot of the **result** for **one portion** (albeit the major portion) of the amendment. It's not the amendment text itself, and it doesn't cover all of the content needed. The Annex to colr_gradients_spec has the actual and complete amendment content that is needed. If you think this is unhelpful, then let's delete this.

PeterConstable:
Removed the .md file.

Issue #204, opened 2021-01-11
abrupt change as radial gradient circle moves from/to contained within/external to other circle


anthrotype:
Consider a COLRv1 variable font with a radial gradient in which the start circle's centre point (c0) moves from being outside to being inside the end circle or viceversa. It would look like the following SVG animation:

https://user-images.githubusercontent.com/6939968/104179575-f7905b00-5403-11eb-9d8a-f27871bafc44.mp4

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="300" height="600">
<defs>
<radialGradient id="grad" cx="50%" cy="50%" r="30%" fx="50%" fy="50%" fr="10%">
	<animate attributeName="fx" values="-2.5%;102.5%;-2.5%" dur="6s" repeatCount="indefinite" begin="0s"/>
    <stop stop-color="#FFDB80" offset="0%"/>
    <stop stop-color="#FF791F" offset="50%"/>
    <stop stop-color="#E09F00" offset="100%"/>
</radialGradient>
</defs>
<rect width="300" height="300" fill="url(#grad)"/>
</svg>
```

So, there is a very stark change as one circle goes from being inside to being outside the other, as you can see. 
It's as if I'm looking at the interior surface of a hollow 3D cone that's pointing away from me; then as the cone (or my point of view) rotates with the tip now lying outside the circle, I find myself outside the cone looking at its exterior surface.
Intuitively I would have tought this imaginary cone to point towards me, and that I'd always look at it from the outside, thus avoiding the abrupt change at the boundary.

In any case, I think the current behavior is intentional and by design, as it appears consistent with the way HTML Canvas createRadialGradient and SVG2 specify radial gradients, as well as the way implementation work in this situation (the animation above is taken from Google Chrome).

However, I was actually not aware of this until I stumbled on a bug in our COLRv1 compiler where, because of some rounding issue, one circle centre that in the input SVG was exactly on the perimeter of the other circle, ended up shifting to being outside in the compiled COLRv1 table, producing an unexpected incorrect rendering (https://github.com/googlefonts/picosvg/issues/158).

Honestly I don't know if this is or would become a problem, I simply wanted to highlight this here in case others think we should address this somehow.

anthrotype:
[...]

Issue #150, opened 2020-12-09
Clarify expected behaviour of radial gradient if circles are the same.


PeterConstable:
If both circles have the same center and the same non-zero radius, then the algorithm would paint only a circle (without thickness, or minimum possible thickness) with the first extended color at positive infinity. "Color at infinity" isn't well defined. The areas within or outside the circles would not be painted.

SVG has a noticeably different behaviour, which resembles having circles of _very slightly different_ radii, with that difference approaching zero. 

* For pad mode, the interior and exterior are painted with solid colors of the first and last (or vice versa?) stops. 

* For repeat and reflect modes, the interior and exterior are painted with the same colour --- for the red/yellow/blue in our example images, it comes out a muddy yellow. This is where the different-radii-approaching-zero behaviour can be seen: if you make the circles with slightly different radii and zoom in, you'll see there is gradation; but zoomed out, the effect looks like a solid muddy yellow but with moirée patterns. If the difference in radii is made smaller, the moirée pattern will change, but the overall color is the same.

svgeesus:
The rendered effect depends a lot on how much sub-pixel sampling is used. It tends to highlight numerical precision issues as well.

PeterConstable:
If I take the SVG we used to generate the first radial gradient example (currently fig 5.13 in the annex), and update circle 0 to have the same center and radius as circle 1, the result in Chromium browsers looks like this;

![image](https://user-images.githubusercontent.com/26614869/101793626-dc57c600-3aba-11eb-9b55-27f056e40f61.png)

In Firefox, it looks like this:

![image](https://user-images.githubusercontent.com/26614869/101793460-b03c4500-3aba-11eb-8dd4-a9d417b7ee86.png)

The Firefox result is consistent with how I would interpret the algorithm, which the Chromium result is not.

But if the radii are just slightly different, then they are the same, and consistent with the algorithm (as expected). Here, the radii are set to 79 and 80:

![image](https://user-images.githubusercontent.com/26614869/101794227-80da0800-3abb-11eb-88a4-41116cba3de3.png)

Making the difference smaller—79.5 vs. 80—changes the Moiré pattern, as I'd expect:

![image](https://user-images.githubusercontent.com/26614869/101794618-e9c18000-3abb-11eb-8252-9bd56ab3dd72.png)

Those are as I'd expect. It's really just the case of circles being exactly the same that's in question.

rsheeter:
[...]

PR #206, opened 2021-01-12
add note for rad grad re sharp transitions if rel'nship of circles vary

PeterConstable:
Cf #204


anthrotype (approved):

OFF_AMD2_WD.md
```diff
@@ -321,8 +321,9 @@ yellow at 0.5, and red at 1.0.
 A radial gradient provides gradation of colors along a cylinder defined by two
 circles. The gradient is defined by circles with center c₀ and radius r₀, and
 with center c₁ and radius r₁, plus a color line. The color line aligns with the
-two circles by associating stop offset 0 with the first circle (with center c₀)
-and aligning stop offset 1.0 with the second circle (with center c₁). 
+two circles by associating stop offset 0 with the circumference of the first
+circle (with center c₀) and aligning stop offset 1.0 with the circumference of
+the second circle (with center c₁).
```
drott, line 326: I think we touched this before when writing the radial gradients chapter, and at the time you suggest to associate the stop offset with "a point on the circumference of the circle". I don't understand why we would want to describe the stop offsets associated with the circumferences here. Maybe my understanding of circumference in English is not fully accurate: Does the circumference describe the "outline" of the circle and its contents, or only the outline? 

In the algorithm, the wording is:

> For all values of ω where r(ω) > 0, starting with the value of ω nearest to positive infinity and ending with the value of ω nearest to negative infinity, draw the circumference of the ellipse resulting from translating a circle with radius r(ω) centered at position (x(ω), y(ω)), with the color at ω, but only painting on the parts of the bitmap that have not yet been painted on by earlier circles in this step for this rendering of the gradient.

Here, circumference means, draw the full circle's area (with the don't-overdraw previous pixels restriction), doesn't it?

I would prefer to keep this paragraph as is - the stop offsets are associated with the circles, the algorithm describes the rest, i.e. drawing the circumferences from positive infinity towards negative infinity etc.

drott (commented):
> Subject: For add note for rad grad re sharp transitions if rel'nship of circles vary

Minor: I personally find abbreviated commit message headers like this very hard to read. "For add note for rad grad" takes me a few seconds to parse. Can we try to avoid abbreviations while still keeping them short?

OFF_AMD2_WD.md
PeterConstable: In English, "circle" is ambiguous (for a mathematician, it's a 1D curve; for an artist, it's a 2D shape), but "circumference" is not: the circumference of a circle is 1 dimensional, i.e., a curve with no thickness.

I was inclined to say that the color line is aligned to the circumferences to avoid confusion with alignment to the centers. But you got me wondering: is there a difference?

The algorithm states, "drawing the circumference", which would _not_ mean drawing the contained region. If that's what were needed, then it should be worded differently to make that clear—e.g., "draw a filled circle".

But I think it turns out that either (i) (conceptually) drawing only the circumferences _or_ (ii) drawing filled circles will produce the same result. For the former, each incremental step contributes to the fill for the next incremental circumference; with the latter, the don't-repaint-pixels constraint means that, on each incremental step, the only pixels drawn are for the circumference.

In any case, I'm fine with reverting the wording in the opening paragraph since the intro doesn't need to provide the specificity, which the algorithm provides.

PeterConstable: Btw, the wording in the algorithm, "draw the **circumference**", is the reason I opened #150: if the circles are the same, the interior region is never painted.

PeterConstable:
I've included text to cover the case of radial gradients defined with close-proximity circles that result in interference patterns when displayed. (In the previous commit, I had it in a note, but ISO guidance says that notes don't include SHOULD text.) This is related to the original issue Cosimo raised since variation in a such a radial gradient definition can cause the gradient to pass through stages with sharp transitions involving rapidly-changing interference patterns.

You can open the new .svg in a browser and just resize the window to cause the graphic to scale and you'll see what I mean. Variation can do the same thing.

OFF_AMD2_WD.md
drott: Thanks for the additional explanations. Yes, now I see, then my understanding of circumference was not wrong - but taking the algorithm into account and how it traverses from +∞ to -∞ then results in the circles, I think I had a misconception there. SGTM re keeping the wording here, thanks.

drott (approved):
LGTM

anthrotype (approved):

OFF_AMD2_WD.md
vlevantovsky: Admittedly, I am not an English expert, but I did study math, a lot of it, and unlike any other subject - I like the precision of math definitions that remove any ambiguity and leave no room to interpretations. 
With this in mind, a circle is ( x - a )^2 + ( y - b )^2 = r^2, where (a, b) is the center and r is the radius, and (x, y) define coordinates of points along a circular line. I.e., a circle is a shape defined by a set of points in а 2D plane with the property that all of the points of that line are on an equal distance r of a fixed point (a, b) of that plane called the center of the circular line. 
On the other hand, the circumference of a circle is a singular scalar number that indicates a length of the circular line, and it does not have a shape - it's just a numerical value. Using the radius of a circle, the formula to calculate circumference is simply this: C = 2πr.
Considering the above, I believe that the right term to use is a circle, not circumference.

PeterConstable: I've removed "circumference" from the opening paragraph.

In the algorithm statement, the wording "draw the circumference" was taken from the HTML spec. You're right that circumference is, in careful usage, a scalar value (0D), while in this usage it's meaning is that of a 1D path—what a mathematician would call a circle. We can change the wording to "draw a circle", though unfortunately readers would read that in two different ways: some will understand that as a 1D path and some will understand it to mean a 2D shape (drawing the fill as well). That may be why the HTML spec chose to use "circumference" in this case: it's not strictly correct, but it's less likely that the intended meaning would be misunderstood.

But IIUC, the algorithm would also produce the same result if it said, "draw the filled circle".

At this point, I'm going to merge this PR. If we want to tweak wording here, let's open another issue.

vlevantovsky: I am not aware of any mathematician who would call a line drawn on a 2D plane as a 1D path - I am having hard time to even imagine what that means - the only reasonable explanation of a 1D path is a singular axis (like a timeline). Like I said, circle is a line in 2D space, just like any other line - it doesn't matter whether it's a straight line defined by y = kx, or something more complex. An algorithm can draw a line, but it cannot "draw the circumference" of a circular line. If one [popular] spec has ambiguous incorrect wording, it's not a good reason to reproduce the same mistake in our spec.

vlevantovsky: If we want to explicitly differentiate a circle from a ring or a disk - I'd suggest to use the term "draw a circular line", but saying "draw a circle" is perfectly fine too.

PeterConstable: Well, I know mathematicians that do. I first encountered that when I was studying math, and it seemed non-intuitive, until you consider that a curve has no thickness. The surface of a sphere is a 2D space since it is a surface and every point can be described using two coordinates.

But I'm not opposed to changing wording. Please open a new issue.

vlevantovsky: Why do we need a new issue for something that we already discussed and agreed on? If you're not opposed to changing the wording - let's just change it and forget about it.

PeterConstable: I've created a new PR with revised wording: #215

Issue #196, opened 2021-01-04
Clarify the description of the linear gradients


vlevantovsky:
In the proposed new sub-clause 5.7.11.1.2.2 "Linear Gradients", the second paragraph is as follows:

> An additional point, p₂, is also used to rotate the gradient orientation in the space on either side of the line defined by p₀ and p₁. The vector from p₀ to p₂ can be referred to as the rotation vector. If the rotation vector is colinear with the line p₀p₁, there is no rotation: colors in the space on either side of the line p₀p₁ extend in the perpendicular direction. But if the rotation vector is not colinear, the gradient is drawn skewed by the angle between p₀p₁ and p₀p₂.

I find the use of the term "colinear" very confusing when it is applied to "rotation vector". The term "colinear" is commonly used to define points that belong to the same line. In our case, when we have three points p₀, p₁, and p₂ that define two vectors  p₀p₁ and p₀p₂ - the fact that points p₀, p₁, and p₂ may be colinear (i.e., belong to the same line) doesn't say anything about _directions_ of the two vectors p₀p₁ and p₀p₂. These vectors may have the same direction when colinear points on the line positioned in sequence (p₀, p₁, p₂) or (p₀, p₂, p₁) , but the same two vectors would have opposite directions if colinear points positioned in sequence (p₂, p₀, p₁) or (p₁, p₀, p₂). I would suggest to rephrase the wording of the second paragraph to clarify that it is the difference in *directions* between two vectors p₀p₁ and p₀p₂ that determine the angle with which the gradient is skewed. 

Also, it is not clear [neither from the description nor from examples that follow] what should be the result of rendering a linear gradient when the skew angle is more than 90 degrees (should the color line direction be reversed?). It may also be worth mentioning what role (if any) the magnitude of the "rotation vector" plays in determining linear gradient rotation. Currently, the description is silent about it, and it appears (as illustrated by figures 5.12 - 5.14) that the magnitude of rotation vector has no effect.

anthrotype:
(I paste this below as a reminder to myself, I'll elaborate more later)

https://github.com/googlefonts/nanoemoji/blob/885d8c6e6410403bc37d67e9aa6803f83419702b/src/nanoemoji/svg.py#L188-L199
```python
    # P2 allows to rotate the linear gradient independently of the end points P0 and P1.
    # Below we compute P3 which is the projection of P1 onto the line between P0 and P2
[...]

PR #207, opened 2021-01-12
For lin grad, clarify effect of rot'n vector if angle is >90

PeterConstable:
This is proposed text to address #196

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -289,11 +289,12 @@ p₁ are interpolated using the color line.
 
 An additional point, p₂, is also used to rotate the gradient orientation in the
 space on either side of the line defined by p₀ and p₁. The vector from p₀ to p₂
-can be referred to as the *rotation vector*. If the rotation vector is colinear
-with the line p₀p₁, there is no rotation: colors in the space on either side of
-the line p₀p₁ extend in the perpendicular direction. But if the rotation vector
-is not colinear, the gradient is drawn skewed by the angle between p₀p₁ and
-p₀p₂.
+can be referred to as the *rotation vector*, although the magnitude and
+direction are not significant. If the the line p₀p₂ is colinear with the line
+p₀p₁, there is no rotation: colors in the space on either side of the line p₀p₁
+extend in the perpendicular direction. But if the line p₀p₂ is not colinear with
+the line p₀p₁, the gradient is drawn skewed by the acute angle (less than 90°)
+between the lines defined by p₀p₁ and by p₀p₂.
 
 If the dot-product (p₁ - p₀) · (p₂ - p₀) is zero (or near-zero for an
 implementation-defined definition) then the gradient is ill-formed and shall not
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -284,20 +284,20 @@ stops, and set the extend mode to *reflect*.
 
 A linear gradient provide gradation of colors along a straight line. The
 gradient is defined by two points, p₀ and p₁, plus a color line, with stop
-offset 0 aligned to p₀ and stop offset 1.0 aligned to p₁. Colors along the line
-passing through p₀ and p₁ are interpolated using the color line. For each
-position along the line, the color at that position is projected on other side
-of the line.
+offset 0 aligned to p₀ and stop offset 1.0 aligned to p₁. (The line passing
+through p₀ and p₁ will be referred to as line p₀p₁.) Colors at each position on
+line p₀p₁ are interpolated using the color line. For each position along line
+p₀p₁, the color at that position is projected on other side of the line.
 
 An additional point, p₂, is also used to rotate the gradient orientation in the
-space on either side of the line defined by p₀ and p₁. The line passing through points p₀ and p₂
-determine the direction in which the gradient progresses. If the the line p₀p₂ is colinear with the line
-p₀p₁, there is no rotation: colors in the space on either side of the line p₀p₁
-extend in the perpendicular direction. But if the line p₀p₂ is not colinear with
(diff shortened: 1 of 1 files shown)


anthrotype:
maybe instead of (or if you like in addition to) rotation and angles, you could mention this alternative way which I find useful to think about a transformed linear gradient, which I hinted at in https://github.com/googlefonts/colr-gradients-spec/issues/196#issuecomment-758817422

Basically, when given the three P0, P1 and P2 you want to find the point (shal we call it P3) that gives you the _effective_ linear gradient between P0 and P3 which is identical and functionally equivalent to the one between P0 and P1 as rotated by P2.

The point P3 is the orthogonal projection of P1 onto the line between P0 and P2.
P0, P1, and P2 describe two vectors: `v = P1 - P0` and `u = P2 - P0`. You compute the [vector projection](https://en.wikipedia.org/wiki/Vector_projection) of `v` onto `u` -- which returns another vector -- and add that to P0 to obtain P3.

PeterConstable:
That wording gets complicated and is a bit hard to follow. But there's a related idea that might be useful: for each position along the colour line (positioned on p0p1), the colour at that position is projected to the sides in a direction orthogonal to the line p0p2.

anthrotype:
yes, I like that. It's also consistent with my algorithm for finding the implied P3 projection

anthrotype:
however, in the end an implementor reading this spec really wants to find a way to reduce these triplet P0, P1 and P2 to two points which describe the start and end position of the final linear gradient, that they can plug in a graphics API that  mostly already knows how to draw a linear gradient between two points..

PeterConstable:
> 
> 
> however, in the end an implementor reading this spec really wants to find a way to reduce these triplet P0, P1 and P2 to two points which describe the start and end position of the final linear gradient, that they can plug in a graphics API that mostly already knows how to draw a linear gradient between two points..

On the one hand, the ability to express an equivalent gradient using only one vector begs the question as to why we specify it using two. On the other hand (and this is a partial response to that question), the vector p0p3 wouldn't give the placement of the colour line: stop offset 0 aligns with p0, but stop offset 1 doesn't align with p3.

rsheeter:
The proposed wording feels like it calls [p0 p2] a vector then strips away it's vector'ness. Describing it in terms of how to compute a point P3 such that you can draw a linear gradient P0=>P3, compute colors along it and project those colors orthogonal to the line, seems like it might be clearer and more helpful to implementers?

Such an explanation will likely demand explanation of why our linear gradient isn't directly defined that way. I believe the answer was to enable variation scenarios where the points tracked other points, such as points in geometry?

> stop offset 1 doesn't align with p3.

you want it to iiuc

PeterConstable:
>Such an explanation will likely demand explanation of why our linear gradient isn't directly defined that way. I believe the answer was to enable variation scenarios where the points tracked other points, such as points in geometry?

Yup, I raised that question before. Others will likely wonder.

I've rev'd the wording, adding a note describing the p0p3 vector as Cosimo suggested, and also anticipating the above question.

OFF_AMD2_WD.md
```diff
@@ -284,16 +284,29 @@ stops, and set the extend mode to *reflect*.
...
 
 An additional point, p₂, is also used to rotate the gradient orientation in the
-space on either side of the line defined by p₀ and p₁. The vector from p₀ to p₂
-can be referred to as the *rotation vector*. If the rotation vector is colinear
-with the line p₀p₁, there is no rotation: colors in the space on either side of
-the line p₀p₁ extend in the perpendicular direction. But if the rotation vector
-is not colinear, the gradient is drawn skewed by the angle between p₀p₁ and
-p₀p₂.
+space on either side of the line defined by p₀ and p₁. The line passing through points p₀ and p₂
+determine the direction in which the gradient progresses. If the the line p₀p₂ is colinear with the line
```
vlevantovsky, line 294: A minor nit for clarity (only because I am not in favor of using the term "colinear" in reference to lines): 

If the points p₀, p₁ and p₂ are colinear, there is no rotation: color in the space on either side of the line p₀p₁
extend in the perpendicular direction. But if the points p₀, p₁ and p₂ are not colinear, the gradient is drawn skewed by the acute angle (less than 90°) between the lines defined by p₀p₁ and by p₀p₂.

PeterConstable: I've revised further to address this concern.

vlevantovsky: Thank you.

rsheeter:
This looks OK to me but I'd like @anthrotype or @drott to double-check as they've spent more time knee deep in the gradients.

rsheeter (approved):

OFF_AMD2_WD.md
```diff
@@ -284,16 +284,29 @@ stops, and set the extend mode to *reflect*.
...
 
 An additional point, p₂, is also used to rotate the gradient orientation in the
-space on either side of the line defined by p₀ and p₁. The vector from p₀ to p₂
-can be referred to as the *rotation vector*. If the rotation vector is colinear
-with the line p₀p₁, there is no rotation: colors in the space on either side of
-the line p₀p₁ extend in the perpendicular direction. But if the rotation vector
-is not colinear, the gradient is drawn skewed by the angle between p₀p₁ and
-p₀p₂.
+space on either side of the line p₀p₁. The line passing through points p₀ and p₂
+(line p₀p₂) determine the direction in which the gradient progresses. If p₂ is
```
anthrotype, line 294: maybe "orientation" would be better here than "direction"? the latter makes me think of an arrow or vector.

anthrotype (approved):
LGTM thanks!

OFF_AMD2_WD.md
anthrotype: never mind, it's ok like this

drott (approved):
LGTM, too.

PR #210, opened 2021-01-13
fix markup of table for PaintColrLayers def'n

PeterConstable:


PeterConstable:
An artifact of wrapping lines at 80 columns—it breaks some markdown markup.

Issue #211, opened 2021-01-13
Specify that compositing is also performed in linear space


drott:
Similar to the discussion in how to interpolate colors, issue #27, we should probably explciclty specify that the compositing operations are to be performed in linear space as well.

PeterConstable:
Changes (#214) merged.

PR #214, opened 2021-01-14
Add CPAL edits re sRGB 'gamma' and interpolation

PeterConstable:
There's a little bit of intro-to-concepts content since readers of the OFF/OT spec come from a wide variety of backgrounds, and jumping straight to a statement like "compute using linear RGB" won't mean anything for many.

Linear RGB for blending isn't added here since that's will fit better in the COLR section on compositing, so will come separately. But a "for more info" pointer to the CPAL section can be included.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2057,7 +2057,8 @@ the non-linear sRGB scale (the default), or using a linear-light scale by
 applying the inverse sRGB transfer function.
 
 For gradient color values in the COLR table, interpolation must be computed
-using linear-light values after applying the inverse sRGB transfer function.
+using linear-light values (i.e., after applying the inverse sRGB transfer
+function).
 
 After an interpolated color value is computed, whether or not the non-linear
 sRGB transfer function needs to be re-applied is determined by the requirements
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2006,7 +2006,7 @@ positions along a color line, with color values for other positions on the color
 line derived by interpolation.
 
 When interpolating color values, linear interpolation between color stop
-positions is used. For example, suppose adjacent color stops are specied for
+positions is used. For example, suppose adjacent color stops are specified for
 positions 0.5 and 0.9 on a color line, and a color value is being calculated for
 position 0.8. The color value of the first color stop will contribute 75% of the
 value ((0.8 - 0.5) / (0.9 - 0.5)), and the color value of the second color stop
@@ -2019,13 +2019,22 @@ colors as well as handling of alpha need to be considered.
 Representations of sRGB color values are expressed as levels of red, green and
 blue color “primaries” with specific, absolute chromaticity values, which are
 defined in the sRGB specification. Color-primary levels can potentially be
-expressed using a linear-light-energy scale. For sRGB, however, standard
-practice is to represent levels using a scale defined by a transfer function,
-sometimes referred to as a non-linear or “gamma” representation. This transfer
-function is also defined in the sRGB specification. (See [CSS Color Module Level
-4, section 10.2](https://www.w3.org/TR/css-color-4/#predefined) for details.) In
-the CPAL table, sRGB color values are always specified in terms of the
-non-linear, sRGB transfer function.
+expressed using a linear scale that correlates directly to light energy. (On a
+linear scale, for example, a doubling of a color value would correspond to a
+doubling of display luminance.) For sRGB, however, standard practice is to
+represent levels using a scale defined by a non-linear transfer function,
+sometimes referred to as “gamma”. This transfer function is also defined in the
+sRGB specification. (See [CSS Color Module Level 4, section
+10.2](https://www.w3.org/TR/css-color-4/#predefined) for details.) In the CPAL
+table, sRGB color values are always specified in terms of the non-linear, sRGB
+transfer function. 
+
+NOTE: An advantage of representing colours using a non-linear scale is that it
+allows more effective use of limited bit depth when color-primary levels are
+represented as integers: smaller differences in light energy can be represented
+for lower levels than for higher levels. This is beneficial since the human
+visual system is more sensitive to differences at low luminance levels than to
+differences at high luminance levels.
 
 When interpolating colors, different results will be obtained if the
 interpolation is computed using the non-linear scale for color levels than if
@@ -2055,21 +2064,21 @@ ColorIndex records in the COLR table, the alpha value from the ColorIndex record
 components as well. Interpolated values are then calculated by linear
 interpolation using these pre-multiplied, linear R, G and B values.
 
-Alpha components use a linear scale and can be directly interpolated apart from
-the R, G and B components. Since the interpolated color values are not stored,
-however, this computation is never required.
+NOTE: Alpha components use a linear scale and can be directly interpolated apart
+from the R, G and B components without any linearlization step. Since the
+interpolated color values are not stored, however, this computation is never
+required.
 
 While color values are specified as 8-bit integers, the interpolation
 computations will require greater precision in each of the linearization,
-pre-multiply, and interpolation steps. These operations should generally be done
-using floating values. Also, when rendered results are to be presented on a
-imaging device with known characteristics, visual bending effects in a gradient
-can be minimized by taking full advantage of the color bit depth supported by
-the device. For instance, if a display supports 10- or 12-bit quantization per
-color channel, then ideally the ramp of color values in a gradient would use
-that level of quantization. Other factors from the presentation context may also
-affect the available capabilities, however. Therefore, no minimum level of
-precision is specified as a requirement.
+pre-multiply, and interpolation steps. Also, when rendered results are to be
+presented on a imaging device with known characteristics, visual banding
+artifacts in a gradient can be minimized by taking full advantage of the color
+bit depth supported by the device. For instance, if a display supports 10- or
+12-bit quantization per color channel, then ideally the ramp of color values in
+a gradient would use that level of quantization. Other factors from the
+presentation context may also affect the available capabilities, however.
+Therefore, no minimum level of precision is specified as a requirement.
 
 ## Changes to OFF 7.2.3 Item variation stores
 
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2057,7 +2057,7 @@ the non-linear sRGB scale (the default), or using a linear scale by applying the
 inverse sRGB transfer function.
 
 For gradient color values in the COLR table, interpolation must be computed
-using linear values.
+using linear values after applying the inverse sRGB transfer function.
 
 After an interpolated color value is computed, whether or not the non-linear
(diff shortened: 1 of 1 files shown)


OFF_AMD2_WD.md
```diff
@@ -1788,6 +1789,93 @@ indicate that an item has no variation data, the index fields shall be set to
...
+will contribute 25% of the value. Interpolated values at each position of the
+color line are computed in this way for each of the R, G and B color components.
+
+When interpolating color values, specific aspects of the representation of
+colors as well as handling of alpha need to be considered.
+
+Representations of sRGB color values are expressed as levels of red, green and
+blue color “primaries” with specific, absolute chromaticity values, which are
+defined in the sRGB specification. Color-primary levels can potentially be
+expressed using a linear-light-energy scale. For sRGB, however, standard
```
drott, line 1827: > linear-light-energy scale
is introduced here without much explanation. Perhaps we can explain a bit what it is: i.e. evenly spaced segmentation of the color channel contribution to the color value, and about how a linear encoding would lead to lack of resolution in the darker parts of the color channel due to how visual perception works?

OFF_AMD2_WD.md
```diff
@@ -1788,6 +1789,93 @@ indicate that an item has no variation data, the index fields shall be set to
...
+interpolation using these pre-multiplied, linear R, G and B values.
+
+Alpha components use a linear scale and can be directly interpolated apart from
+the R, G and B components. Since the interpolated color values are not stored,
+however, this computation is never required.
+
+While color values are specified as 8-bit integers, the interpolation
+computations will require greater precision in each of the linearization,
+pre-multiply, and interpolation steps. These operations should generally be done
+using floating values. Also, when rendered results are to be presented on a
```
drott, line 1870: I suggest to drop the recommendation to use float values, this seems too specific. As discussed in the meeting yesterday, perhaps we can say instead that it should be done with higher precision than with the 8bit values coming from the encoding?

OFF_AMD2_WD.md
```diff
@@ -1788,6 +1789,93 @@ indicate that an item has no variation data, the index fields shall be set to
...
+imaging device with known characteristics, visual bending effects in a gradient
+can be minimized by taking full advantage of the color bit depth supported by
```
drott, line 1872:
```suggestion
imaging device with known characteristics, visual banding effects in a gradient
can be minimized by taking full advantage of the color bit depth supported by
```

OFF_AMD2_WD.md
```diff
@@ -1788,6 +1789,93 @@ indicate that an item has no variation data, the index fields shall be set to
...
+ColorIndex records in the COLR table, the alpha value from the ColorIndex record
+(with variation, in a variable font) is multiplied into the R, G and B
+components as well. Interpolated values are then calculated by linear
+interpolation using these pre-multiplied, linear R, G and B values.
+
+Alpha components use a linear scale and can be directly interpolated apart from
+the R, G and B components. Since the interpolated color values are not stored,
+however, this computation is never required.
+
+While color values are specified as 8-bit integers, the interpolation
```
drott, line 1867: Move this paragraph to a note?

drott (commented):
Thanks, Peter, for drafting this chapter, a few comments below. Disclaimer: I am certainly not an expert on color and related computations, so I might confuse things as well here.

OFF_AMD2_WD.md
PeterConstable: This stuff is actually deeply complicated when you get into the details. 

(The linear RGB values in an image are representations captured illuminance and expected display luminance. The latter is related to radiance by incorporation of the standard observer spectral responsiveness in spectral integration, and is measured in watts per square meter per steradian... and on and on... And the background for non-linear transfer functions also gets into the response behaviour in electron guns and phosphors in CRTs, as well as factors related to expecting viewing environments... and on and on...)

So a bit of thinking is needed to figure what can be said that's accurate, understandable, and useful.

The distribution of quantized distinctions and the nature of human visual response (more sensitive to differences at low luminance levels than at high levels) is probably useful. So, I'll work on what to say that's also accurate and understandable.

PeterConstable:
I've added the text on blending using linear color values (#211).

OFF_AMD2_WD.md
PeterConstable: Yeah, I almost didn't include that.

OFF_AMD2_WD.md
```diff
@@ -1836,6 +1843,102 @@ indicate that an item has no variation data, the index fields shall be set to
...
+computations will require greater precision in each of the linearization,
+pre-multiply, and interpolation steps. Also, when rendered results are to be
+presented on a imaging device with known characteristics, visual banding
+artifacts in a gradient can be minimized by taking full advantage of the color
+bit depth supported by the device. For instance, if a display supports 10- or
+12-bit quantization per color channel, then ideally the ramp of color values in
+a gradient would use that level of quantization. Other factors from the
+presentation context may also affect the available capabilities, however.
+Therefore, no minimum level of precision is specified as a requirement.
```
drott, line 1940: Nit from a non-native English speaker, so this might be just a personal impression: The trailing ", however." irritates my reading flow a bit when the next word is "Therefore...". I suggest to move "however" to the beginning of the sentence or after _may_.

drott (commented):
This looks good to me but it'd be great to have @svgeesus 's feedback on this PR. Also, @bungeman , do you have any further comments?

svgeesus:
> This looks good to me but it'd be great to have @svgeesus 's feedback on this PR

Yes I will try to review today

PeterConstable:
@svgeesus Any chance you can review this soon?

OFF_AMD2_WD.md
```diff
@@ -1831,13 +1831,20 @@ The compositionMode value must be one of the values defined in the CompositeMode
...
+the source is composited or blended into the backdrop using the specified
+composite mode. The supported modes are taken from the W3C [Compositing and
+Blending Level 1][1] specification. Details on each mode, including
+specifications of the required calculations using pixel color and alpha values,
+are provided in that specification.
+
+While color values obtained from the CPAL table are represented in sRGB using
+the non-linear transfer function defined in the sRGB specification, the
+compositing and blending calculations are done after applying the inverse
+transfer function to derive linear RGB values. For more information regarding
```
svgeesus, line 1845: Linear is being used in two senses here - linear interpolation, and linear-light. I suggest therefore changing this from linear RGB""to "linear-light sRGB"

OFF_AMD2_WD.md
```diff
@@ -1972,6 +1979,104 @@ indicate that an item has no variation data, the index fields shall be set to
...
+When interpolating colors, different results will be obtained if the
+interpolation is computed using the non-linear scale for color levels than if
+using the linear scale. For interoperable results, whether the non-linear or
+linear scale is to be used needs to be specified.
+
+For gradient color values in the SVG table, the required interpolation behavior
+is defined in the SVG 1.1 specification: the [‘color-interpolation’
+property](https://www.w3.org/TR/SVG11/painting.html#ColorInterpolationProperty)
+can be used in an SVG document to declare whether interpolation is done using
+the non-linear sRGB scale (the default), or using a linear scale by applying the
```
svgeesus, line 2045: that is the default, sadly, but not because it is better necessarily. It was also faster (in 1999).

(not a change request, just an observation)

OFF_AMD2_WD.md
```diff
@@ -1972,6 +1979,104 @@ indicate that an item has no variation data, the index fields shall be set to
...
+inverse sRGB transfer function.
+
+For gradient color values in the COLR table, interpolation must be computed
+using linear values after applying the inverse sRGB transfer function.
```
svgeesus, line 2049: Again, ""linear-light"would be clearer here

OFF_AMD2_WD.md
```diff
@@ -1972,6 +1979,104 @@ indicate that an item has no variation data, the index fields shall be set to
...
+of the implementation context.
+
+For both the COLR and SVG tables, interpolation must be done with alpha
+pre-multiplied into each linearized R, G and B component. For alpha specified in
+a CPAL ColorRecord, the value is converted to a floating value in the range [0,
+1.0] by dividing by 255, then multiplied into each R, G and B component. For
+ColorIndex records in the COLR table, the alpha value from the ColorIndex record
+(with variation, in a variable font) is multiplied into the R, G and B
+components as well. Interpolated values are then calculated by linear
+interpolation using these pre-multiplied, linear R, G and B values.
```
svgeesus, line 2062: This is correct as far as it goes, but misses a step. Add:

Once interpolation of the premultiplied red, green and blue values, and the alpha, is complete the results are then un-premultiplied by dividing each of the premultiplied red, green and blue values by the value of alpha.

svgeesus (changes requested):
Overall very good and requiring minimal changes.

The un-premultiplication step to get back actual colors was the major missing portion.

OFF_AMD2_WD.md
PeterConstable: @svgeesus: Part of your proposed addition isn't clear to me: "dividing... by the value of alpha". Which alpha value? 

I'm guessing that, for each stop offset of a gradient, the interpolated, premultiplied RGB values are divided by the interpolated alpha? I've updated the text assuming that to be the case. Please confirm.

svgeesus: Yes, the interpolated, premultiplied color value is divided by the interpolated alpha value, to get back actual colors.

OFF_AMD2_WD.md
```diff
@@ -2035,18 +2035,18 @@ differences at high luminance levels.
...
 property](https://www.w3.org/TR/SVG11/painting.html#ColorInterpolationProperty)
 can be used in an SVG document to declare whether interpolation is done using
-the non-linear sRGB scale (the default), or using a linear scale by applying the
-inverse sRGB transfer function.
+the non-linear sRGB scale (the default), or using a linear-light scale by
+applying the inverse sRGB transfer function.
 
 For gradient color values in the COLR table, interpolation must be computed
-using linear values after applying the inverse sRGB transfer function.
+using linear-light values after applying the inverse sRGB transfer function.
```
svgeesus, line 2049: Actually that is now slightly ambiguous. Perhaps

"must be computed using linear values after (i.e., applying the inverse sRGB transfer function)."

OFF_AMD2_WD.md
```diff
@@ -2062,9 +2062,12 @@ components as well. Interpolated values are then calculated by linear
...
 
 NOTE: Alpha components use a linear scale and can be directly interpolated apart
-from the R, G and B components without any linearlization step. Since the
-interpolated color values are not stored, however, this computation is never
-required.
+from the R, G and B components without any linearlization step.
+
+Once interpolation of the pre-multiplied red, green and blue values and of the
+alpha value is complete, the red, green and blue results are then
+un-premultiplied by dividing each interpolated value by the corresponding
```
svgeesus, line 2069: yes

svgeesus (approved):
One minor wording suggestion but otherwise this is great.

PR #215, opened 2021-01-15
Revise wording for radial gradient algorithm

PeterConstable:
This is to address the issue Vlad raised in #206 that the term "circumference" was being used incorrectly.

The wording "draw the circumference" was taken from the HTML Canvas spec. I'm guessing the intent in using that wording was to convey that what is drawn in the circular path but not the interior region, and that "draw the circle" might be misunderstood to include the interior region. In this initial commit, I've proposed the wording "draw the circular path" to retain that meaning.


vlevantovsky (approved):
Looks good, thank you Peter!

drott (approved):
LGTM, FWIW I am fine with "circle" or "circular line".

Issue #84, opened 2020-10-15
Conic gradients


davelab6:
https://twitter.com/LeaVerou/status/1315655737608146945?s=19

Possible v2 feature?

rsheeter:
Very cool. +1 to v2 unless anyone emerges claiming they simply MUST have it to make use of v1.

behdad:
Honestly I'd put them in since software needs to be updated...

svgeesus:
I would certainly encourage adding Conic gradients to V2. The gradient line is simply a circle. [CanIUse conic gradient](https://caniuse.com/css-conic-gradients)

PeterConstable:
In F2F discussion with Rod, Dominik, Cosimo, deciding to incorporate this now, in COLR v1.

PeterConstable:
This has been added now for COLR v1, so closing.

PR #217, opened 2021-01-15
Add PaintSweepGradient

PeterConstable:
This pertains to #84. There are more changes that would be needed, but I want to first see whether there's interest in incorporating this at this time.


PeterConstable:
To decide on the format, I looked at sweep/conic gradients in Skia and in CSS. (This type isn't supported in SVG2 or in HTML Canvas.) Expressing the placement of the color line using start/stop angles seems the most flexible and amenable to variations. I chose the name "sweep" in part to avoid confusion with our radial gradient, which often takes the appearance of a truncated cone and which Skia calls "two-point conical"; and because it is called "sweep" in Skia.

OFF_AMD2_WD.md
```diff
@@ -503,6 +503,38 @@ caution if the circles are in close proximity (either in a static design or for
...
+
+![A conic gradient](images/colr_conic_gradient.png)
+
+**Figure 5.x Conic gradient**
+
+NOTE: In some contexts, this type of gradient is referred to as a “conic” gradient.
+
+A sweep gradient is defined by a center point, starting and ending angles, and a color line. The angles are expressed in counter-clockwise degrees from the direction of the y-axis on the design grid.
+
+The color line is aligned to a circular path around the center point, with arbitrary radius, with stop offset 0 aligned with the starting angle, and stop offset 1 aligned with the ending angle. The color line progresses in the counter-clockwise direction; for example, if the start and stop angles are both 0°, then stop offset 0.1 is at 36° counter-clockwise from the direction of the y-axis. The color line may be defined using color stops outside the range [0, 1], but only color values in the range [0, 1] are used. In a conic gradient, the extend mode specified for the color line is ignored.
```
drott, line 520:
```suggestion
The color line is aligned to an arc around the center point, with arbitrary radius, with stop offset 0 aligned with the starting angle, and stop offset 1 aligned with the ending angle. The color line progresses in the counter-clockwise direction; for example, if the start and stop angles are both 0°, then stop offset 0.1 is at 36° counter-clockwise from the direction of the y-axis. The color line may be defined using color stops outside the range [0, 1], but only color values in the range [0, 1] are used. In a conic gradient, the extend mode specified for the color line is ignored.
```

OFF_AMD2_WD.md
```diff
@@ -280,6 +283,13 @@ reflected gradient pattern, divide 1 by two times the number of desired
 repetitions, use the result as the maximum stop offset for specified color
 stops, and set the extend mode to *reflect*.
 
+NOTE: While color stops can be defined for stop offsets outside the range [0, 1]
+and the color line is extended indefinitely in either direction, for a sweep
+gradient, only color values within the range [0, 1] are used. The extend mode is
+relevant for a sweep gradient only if color stops are defined for a sub-range
+smaller than [0, 1] and the extend mode is needed to fill in values for the
```
rsheeter, line 290: @drott is that true? - I have hazy memory of discussion of it being fine to use values outside [0, 1]

OFF_AMD2_WD.md
```diff
@@ -577,15 +595,16 @@ The PaintGlyph table has a field for the glyph ID, plus an offset to a child
...
-PaintRadialGradient, can be used as the child paint table. This is illustrated
-in figure 5.29: a PaintGlyph table has a glyph ID for an outline in the shape of
-a triangle, and it links to a child PaintLinearGradient table. The combination
-is used to represent a triangle filled with the linear gradient.
+Any of the basic fill formats—PaintSolid, PaintLinearGradient,
+PaintRadialGradient, or PaintSweepGradient—can be used as the child paint table.
+This is illustrated in figure 5.33: a PaintGlyph table has a glyph ID for an
+outline in the shape of a triangle, and it links to a child PaintLinearGradient
+table. The combination is used to represent a triangle filled with the linear
+gradient.
```
rsheeter, line 603: I worry this language [both before and after this edit] implies *only* the basic formats can be used when in fact there is no such limitation.

OFF_AMD2_WD.md
```diff
@@ -1833,27 +1855,28 @@ function renderPaint(paint)
...
         call renderPaint() passing the paint table referenced by the base glyph ID
 
-    if format 7: // PaintTransformed
-    or if format 8: // PaintTranslate
-    or if format 9: // PaintRotate
-    or if format 10: // PaintSkew
+    if format 8: // PaintTransformed
+    or if format 9: // PaintTranslate
+    or if format 10: // PaintRotate
+    or if format 11: // PaintSkew
```
rsheeter, line 1873: (Optional) for pseudocode I think if format in 8, 9, 10, 11 might be ok

rsheeter:
Initial impression is this looks good to me. I did get to wondering about colorline [0,1] aligning to start/end angle vs being pegged to 0/360 to facilitate varying the angle with a stable colorline (effectively revealing portions of the gradient). Not sure that matters much in reality.

OFF_AMD2_WD.md
PeterConstable: The color line can be defined with stops outside [0,1]; this is called out:

>The color line may be defined using color stops outside the range [0, 1], but
only color values in the range [0, 1] are used.

OFF_AMD2_WD.md
PeterConstable: The paragraph after the explains how the child is not limited to only the basic formats. But that could be spelled out more clearly.

PeterConstable: Actually, the paragraph does spell that out, but re-ordered of the statements would probably help.

OFF_AMD2_WD.md
PeterConstable: Revised

PeterConstable:
> I did get to wondering about colorline [0,1] aligning to start/end angle vs being pegged to 0/360 to facilitate varying the angle with a stable colorline (effectively revealing portions of the gradient). Not sure that matters much in reality.

That is a different approach we could perhaps take. I hadn't considered it since it's not how either Skia or CSS work.

If the colorline [0,1] is always aligned to 0/360, then varying the start or end angles would result in varying the start or end colors. I could see that being desirable in some cases. But if the colorline [0,1] is always aligned to the start/end angles, then the angles could vary while keeping the colors constant. I could see that also being desirable in some cases.

If one wanted to achieve the former effect using formats defined in the latter way, I think you could define one sweep gradient with start/end angles of 0/360, then define a second sweep gradient with different start/end angles, and then use the latter to mask the former.

However, I think achieving the latter effect if the colorline were always aligned to 0/360 would be much harder. Variation can change stop offsets but can't change the color index of a stop or the number of stops. Consider getting the gradient in figure 5.30 to rotate: if the red stop passes through 0/360, then it has to vary from stop offset .833 to 1.0 and then make a discontinuous jump to stop offset 0. Setting up deltas for discontinuous jumps like that is a pain, and in many cases substitution (conditioned by variation settings) to a new glyph ID is considered the better alternative.

Even more generally, I think pinning the colorline [0,1] to 0/360 would be less convenient, and figure 5.30 illustrates this: instead of defining it with stops at 0 and 1, you'd define it for more than one full sweep, with stops at -0.167 and 1.167. (The alternative would require at least 4 stops: orange at 0, yellow at 0.167, red or 0.833, orange at 1; and forget about rotating the colorline, since orange and every other interpolated color would need to be added in CPAL—hugely messy!) It's certainly doable, but less intuitive.

So, for these reasons, and for better alignment with Skia APIs, I think it's better to have colorline [0,1] aligned to the start/end angles, as currently drafted.

rsheeter:
> So, for these reasons, and for better alignment with Skia APIs, I think it's better to have colorline [0,1] aligned to the start/end angles, as currently drafted.

Agreed, I don't suggest any change, just thinking "aloud"

rsheeter (approved):

PeterConstable:
I'd like to get approval from one or both of @drott, @anthrotype before merging.

OFF_AMD2_WD.md
```diff
@@ -187,22 +188,24 @@ In version 1, a solid color fill is specified using a PaintSolid table. See
 
 **5.7.11.1.2 Gradients**
 
-COLR version 1 supports two types of gradients: linear gradients, and radial
-gradients. Both types of gradient are defined using a color line.
+COLR version 1 supports three types of gradients: linear gradients, radial
+gradients, and sweep gradients. Each type of gradient are defined using a color
```
drott, line 192:
```suggestion
gradients, and sweep gradients. Each type of gradient is defined using a color
```

or "All types of gradients are..."

OFF_AMD2_WD.md
drott: Yes. Maybe to make it clearer, we can say

> for a sweep gradient, only color values within the range [0, 1] are painted.

drott: On the other hand, reading below, this is a bit repetitive when adding it to the Color Line section. I'd be in favour of only talking about the [0, 1] mapping in the Sweep gradient definition and dropping it here (or there), but only keep one note about the [0, 1] restricted mapping for sweep gradients.

OFF_AMD2_WD.md
```diff
@@ -516,32 +596,32 @@ The PaintGlyph table has a field for the glyph ID, plus an offset to a child
...
-table is not limited to only the basic fill formats. In general, the child can
-be the root of a sub-graph that describes some graphic composition that
-comprises the fill for the shape. Or, in the alternate view, the glyph outline
-defines a clip region that is applied to the composition.
+The child of a PaintGlyph table is not, however, limited to one of the basic
+fill formats. Rather, the child can be the root of a sub-graph that describes
+some graphic composition that is used as a fill. Another way to describe the
+relationship between a PaintGlyph table and its child sub-graph is that the
+glyph outline specified by the PaintGlyph table defines a bounds, or *clip
+region*, that is applied to the fill composition defined by the child sub-graph.
```
drott, line 615: Could you explain why this change is here? It looks unrelated to PaintSweepGradient to me and thus I'd prefer to have that in a separate pull request.

OFF_AMD2_WD.md
```diff
@@ -502,6 +512,76 @@ caution if the circles are in close proximity (either in a static design or for
...
+**5.7.11.1.2.4 Sweep gradients**
+
+A sweep gradient provides a gradation of colors that sweep around a center
+point. For a given color on a color line, that color projects as a ray from the
+center point in a given direction. This is illustrated in figure 5.29.
+
+NOTE: The following figures illustrate sweep gradients clipped to a circular
+region. Sweep gradients are not bounded, however, and fill the entire space.
+
+![A sweep gradient](images/colr_conic_gradient.png)
```
drott, line 524: Thank you for creating these example images and explaining these details. Minor comment in this regard: How about using as much as possible the same "tonality" / color values as in the other examples? Here, the colour look more like saturated RGB primary colors, in the other example images we used the slightly toned, slightly desaturated colours.

drott (approved):
Thank you Peter for phrasing this and renumbering everything. LGTM with a slight preference to talk about the restriction the [0, 1] range only once in the sweep gradient section. Also a minor request regarding palette values in the examples.

OFF_AMD2_WD.md
PeterConstable: It's here because Rod commented on that in this PR. But I can move it to a different PR.

drott:
Thank you for taking the time to update the images, much appreciated!

PR #219, opened 2021-01-23
Revise linear gradient definition to make p2 the 'normal'

PeterConstable:
Per F2F discussion, changing the semantics for PaintLinearGradient so that P2 defines the direction in which colours are projected to either side of p0p1 instead of the direction in which the gradation progresses.

The reason for the change is that, if p2 is colinear with p0p1 (a common scenario), then it can be problematic for implementations to apply transforms to the gradient and get the expected result. (You should be able to apply the transform to p0, p1, p2 and then paint, rather than having to paint a bitmap and then morph the pixels.) In particular, a skew _should_ affect the gradient, but if p2 is in line p0p1, it will continue to be line p0p1 after the skew, and will not be affected by the skew. 

For example, suppose p0, p1 and p2 are on the x axis; the each color projects in the direction of the y axis. Now suppose a horizontal skew (in x-direction) is applied. Line p0p1 is not affected, but we would want the gradient to be skewed—i.e., the direction in which a given color projects above or below the x axis should be skewed. But in this case, p2 is also not affected, and there is no skew to the gradient. By ensuring that p2 is off p0p1 (except for the degenerate case), that will mean that applying transforms to p0, p1 and p2 will always have the desired effect.


drott:
I need some more time to form an opinion about this while I work through the transforms issues between Skia and FT. Do we expect effects on the file size from this change? @anthrotype? Was there still a way to have P2 optional or zeroed, or did nanoemoji + fonttools previously always put p2 = p2?

anthrotype:
we'd always put P2 == P1 when there was no transform

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,56 @@ extend modes and do not support repetition of gradient patterns. See
...
-aligning stop offset 1.0 to p₃, with each color projecting on either side of
-that line in an orthogonal direction. This specification uses three points, p₀,
-p₁ and p₂, as that provides greater flexibility in controlling the placement and
-orientation of the gradient, as well as variations thereof.
-
-If the dot-product (p₁ - p₀) · (p₂ - p₀) is zero (or near-zero for an
-implementation-defined definition) then the gradient is ill-formed and shall not
-be rendered.
+orthogonal projection of vector p₀p₁ onto a line perpendicular to line p₀p₂ to
+obtain point p₃. The linear gradient defined using p₀, p₁ and p₂ as described
```
anthrotype, line 315: > a line perpendicular to line p₀p₂

... and passing through p0

anthrotype:
now for the simple case of a non-skewed linear gradient a compiler would have to find a place for P2 along the normal of vector P1-P0. The magnitude of vector P2-P0 doesn't matter, only the position of P2 relative to P0 which determines the orientation of the linear gradient same-color lines. I guess I would use a unit vector of magnitude 1 by default.

however because of the need to round floats to integers, the actual rounded coordinates of such P2 could be slightly off compared to the true unrounded P2 (at 1 unit distance along a line orthogonal to P1-P0) and may end up producing a slight (perhaps imperceptible?) tilt of the final linear gradient defined by this triplet P0, P1 and P2 -- even despite the original source linearGradient only defined a p0 and a p1 and no gradient transform.

We could, as I think Peter suggested in one of our last conversations, say that if P2 == P0, then it means that the linear gradient is simple and un-skewed, _as if_ P2 was placed along a line that is orthogonal to P1-P0.

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,56 @@ extend modes and do not support repetition of gradient patterns. See
...
+passes through that position on line p₀p₁ and that is parallel to line p₀p₂ will
+have the color for that position on line p₀p₁.
+
+Points p₁ and p₂ shall not be the same as point p₀. 
+
+If line p₀p₂ is parallel to line p₀p₁ (or the angle between them is near-zero
+for an implementation-defined threshold), then the gradient is ill-formed and
+shall not be rendered.
 
 NOTE: An implementation can derive a single vector, p₀p₃, by computing the
```
anthrotype, line 313:
```suggestion
NOTE: An implementation can derive a single vector, p₃p₀, by computing the
```

when talking about "vectors", and you say p₀p₃, I think about the vector produced by subtracting the two points (P0 - P3), thus an arrow that starts at P3 and points in direction of P0.
I think you mean the opposite. A gradient vector starts at P0 and goes towards P1 (or the effective P3).

OFF_AMD2_WD.md
PeterConstable: fixed

OFF_AMD2_WD.md
PeterConstable: Interesting that we have different intuitions: I think of "p0p3" as going from p0 to p3. Revising wording...

anthrotype: I'm not necessarily saying that mine is the correct one.. If we could put an arrow on top of the notation "p0p3" then it would be clearer.

anthrotype: the revised wordings look good to me, thanks

PeterConstable:
> now for the simple case of a non-skewed linear gradient a compiler would have to find a place for P2 along the normal of vector P1-P0....
> 
> We could, as I think Peter suggested in one of our last conversations, say that if P2 == P0, then it means that the linear gradient is simple and un-skewed, _as if_ P2 was placed along a line that is orthogonal to P1-P0.

Another option (did Rod suggest this?) would be to specify an angle rather than a position. This would entail re-visiting the issue Vlad raised (can the angle be >90°?)

images/linear_gradients_1.svg
```diff
@@ -0,0 +1,52 @@
+<svg viewBox="-301 -206 802 1271" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" height="700">
```
drott, line 1: Minor: We can spell it out in three files, but I also liked the fact that the HTML version that modifies the SVG had the dynamics for moving p2 and the rest automatically following. Would it make sense to still have that dynamic example? On the other hand it did require the additional matrix library. Opinions? 

In any case, we only need one or the other, I think - or if we keep both, the HTML version code should be fixed.

images/linear_gradients_1.svg
```diff
@@ -0,0 +1,52 @@
...
+    <use href="#filled-rect" fill="url(#lingrad-pad)" />
+    <g id="points" fill="none" stroke="#1a2127" stroke-width="2">
+      <circle id="p0dot" cx="0" cy="0" r="4" fill="#1a2127" />
+      <text id="p0label" class="label" x="-40" y="40">p₀</text>
+      <circle id="p1dot" cx="250" cy="0" r="4" fill="#1a2127" />
+      <text id="p1label" class="label" x="250" y="40">p₁</text>
+      <circle id="p2dot" cx="100" cy="100" r="4" fill="#1a2127" />
+      <text id="p2label" class="label" x="60" y="140">p₂</text>
+      <line id="p0p1line" x1="0" y1="0" x2="250" y2="0" stroke-width="3" />
+      <line id="p0p2line" x1="-50" y1="-50" x2="150" y2="150" stroke-width="2" stroke-dasharray="4" vector-effect="non-scaling-stroke" />
```
drott, line 39: I see you're using the same style of the dotted line as in the HTML/SVG example, but somehow in the images it appears slightly thicker to my eye, and the dot for P2 is less visible. Could you rasterize it in Chrome, or change the stroke-width to be a little smaller? (Ditto for the other images.)

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,57 @@ extend modes and do not support repetition of gradient patterns. See
...
 
 Figures 5.12 – 5.14 illustrate linear gradients using the three different color
 line extend modes. Each figure illustrates linear gradients with two different
 rotation vectors. In each case, three color stops are specified: red at 0.0,
 yellow at 0.5, and red at 1.0.
 
-![Linear gradients with different rotation vectors using the pad extend mode.](images/colr_linear_gradients_pad.png)
+![Linear gradients with different color-projection orientations using the pad extend mode.](images/colr_linear_gradients_pad.png)
+
+**Figure 5.12 Linear gradients with different color-projection orientations using the pad extend mode.**
```
drott, line 330: On this one the 0 of P0 is occluded by the dotted line, could you move that out of the way?

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,57 @@ extend modes and do not support repetition of gradient patterns. See
...
 
-**Figure 5.12 Linear gradients with different rotation vectors using the pad extend mode.**
+![Linear gradients with different color-projection orientations vectors using the repeat extend mode.](images/colr_linear_gradients_repeat.png)
 
-![Linear gradients with different rotation vectors using the repeat extend mode.](images/colr_linear_gradients_repeat.png)
+**Figure 5.13 Linear gradients with different color-projection orientations vectors using the repeat extend mode.**
```
drott, line 334: Ditto.

OFF_AMD2_WD.md
```diff
@@ -1511,12 +1516,12 @@ For information about applying a fill to a shape, see 5.7.11.1.3.
...
-| VarFWord | x1 | End point x coordinate. |
-| VarFWord | y1 | End point y coordinate. |
-| VarFWord | x2 | Rotation vector end point x coordinate. |
-| VarFWord | y2 | Rotation vector end point y coordinate. |
+| VarFWord | x0 | Start point (p₀) x coordinate. |
+| VarFWord | y0 | Start point (p₀) y coordinate. |
+| VarFWord | x1 | End point (p₁) x coordinate. |
+| VarFWord | y1 | End point (p₁) y coordinate. |
+| VarFWord | x2 | Orientation end point (p₂) x coordinate. |
+| VarFWord | y2 | Orientation end point (p₂) y coordinate. |
```
drott, line 1524: IMO it's not really an end point - as nothing ends there. Perhaps "Rotation point P2 x/y coordinate"? Mildly prefer rotation, as orientation rings a little confusing to me: it sounds to me like it could mean swapping the color line.

drott (commented):
Text mostly looks good to me, with a suggestion and question whether we want to keep speaking of dot product. 

Some comments on the images.

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,57 @@ extend modes and do not support repetition of gradient patterns. See
...
-
-NOTE: An implementation can derive a single vector, p₀p₃, by computing the
-orthogonal projection of vector p₀p₁ onto line p₀p₂ to obtain point p₃. The
-linear gradient defined using p₀, p₁ and p₂ as described above is functionally
+(line p₀p₂) determines the direction in which colors are projected on either
+side of the color line. That is, for each position on line p₀p₁, the line that
+passes through that position on line p₀p₁ and that is parallel to line p₀p₂ will
+have the color for that position on line p₀p₁.
+
+Points p₁ and p₂ shall not be the same as point p₀.
```
rsheeter, line 307: Suggest instead defining what happens when this is true, such as gradient is degenerate and nothing should be rendered (as in p0p2 p0p1 parallel section)

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,57 @@ extend modes and do not support repetition of gradient patterns. See
...
+(line p₀p₂) determines the direction in which colors are projected on either
+side of the color line. That is, for each position on line p₀p₁, the line that
+passes through that position on line p₀p₁ and that is parallel to line p₀p₂ will
+have the color for that position on line p₀p₁.
+
+Points p₁ and p₂ shall not be the same as point p₀. 
+
+If line p₀p₂ is parallel to line p₀p₁ (or the angle between them is near-zero
+for an implementation-defined threshold), then the gradient is ill-formed and
+shall not be rendered.
```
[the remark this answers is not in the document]
rsheeter: To offer another option, we can just say near-parallel for an implementation defined definition of near

OFF_AMD2_WD.md
rsheeter: Would it make sense to borrow the "normal" terminology of https://www.w3.org/TR/SVG2/pservers.html#Gradients? - we can then cite that if necessary.

OFF_AMD2_WD.md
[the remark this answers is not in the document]
PeterConstable: Weill use near-parallel

images/linear_gradients_1.svg
PeterConstable: The HTML version did not work for me, which is why I created the SVGs. I prefer not requiring any additional libraries to make the example work.

images/linear_gradients_1.svg
PeterConstable: Re-doing the linear gradient images, moving p0 on the left side and rasterizing in Chrome.

OFF_AMD2_WD.md
PeterConstable: Re-doing the linear gradient images, moving p0 on the left side and rasterizing in Chrome.

OFF_AMD2_WD.md
PeterConstable: "Normal" in math mean 'orthogonal'. If we use "normal", then it would beg the question, "Orthogonal to what?" In SVG, they don't have any equivalent to P2, and the "normal" is always orthogonal to p0p1. In our case, depending on placement of P2, p0p2 could be orthogonal, but also could not be. So, when we've referred to it in our discussions as the "normal", that's convenient for our purposes but not really accurate. 

"Iso-color line" would be an accurate description, though that's a neologism I came up with that probably nobody has ever heard of, and only Greek geeks would understand.

PeterConstable: I've added a note in 5.7.11.1.2.2 introducing "rotation" wording, and updated the table in 5.7.11.2.5.4.

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,63 @@ extend modes and do not support repetition of gradient patterns. See
...
 rotation vectors. In each case, three color stops are specified: red at 0.0,
 yellow at 0.5, and red at 1.0.
 
-![Linear gradients with different rotation vectors using the pad extend mode.](images/colr_linear_gradients_pad.png)
+![Linear gradients with different color-projection orientations using the pad extend mode.](images/colr_linear_gradients_pad.png)
+
+**Figure 5.12 Linear gradients with different color-projection orientations using the pad extend mode.**
 
-**Figure 5.12 Linear gradients with different rotation vectors using the pad extend mode.**
+![Linear gradients with different color-projection orientations vectors using the repeat extend mode.](images/colr_linear_gradients_repeat.png)
```
drott, line 338: Since we're talking about a rotation point and rotation vector in line 307 and this was a rotation vector in the previous text - would it make sense to say rotation vector, or just rotation here?

```suggestion
![Linear gradients with different rotation vectors using the repeat extend mode.](images/colr_linear_gradients_repeat.png)
```

OFF_AMD2_WD.md
```diff
@@ -289,52 +289,63 @@ extend modes and do not support repetition of gradient patterns. See
...
-**Figure 5.13 Linear gradients with different rotation vectors using the repeat extend mode.**
+![Linear gradients with different color-projection orientations vectors using the reflect extend mode.](images/colr_linear_gradients_reflect.png)
 
-![Linear gradients with different rotation vectors using the reflect extend mode.](images/colr_linear_gradients_reflect.png)
+**Figure 5.14 Linear gradients with different color-projection orientations vectors using the reflect extend mode.**
 
-**Figure 5.14 Linear gradients with different rotation vectors using the reflect extend mode.**
+When a linear gradient is combined with a transformation (see 5.7.11.1.5), the
+appearance shall be the same as if the gradient were defined using the
+transformed positions of points p₀, p₁ and p₂.
```
drott, line 348: 1) Minor: Did an earlier version of this have this "lemma" as a note? As mentioned, I prefer to phrase this as a note as it already follows from `PaintLinearGradient` and `PaintTransformed`.

2) I find it surprising that this is here after the images. I would find it more suitably placed in line ~328.

images/linear_gradients_1.svg
drott: Ok, please delete the HTML version in this PR then.

drott (approved):
LGTM

OFF_AMD2_WD.md
PeterConstable: I'll change to simply "rotations"

PR #222, opened 2021-01-25
describe transforms applied to radial or sweep gradients

PeterConstable:
Similar wording for linear gradients has been included in PR #219

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -469,13 +469,15 @@ extend mode. Figures 5.24 – 5.26 illustrate this for the different extend mode
 
 NOTE: When a radial gradient is combined with a transformation (see 5.7.11.1.5),
 the appearance will be the same as if the geometry of the two circles were
-transformed and step 3 of the algorithm were performed by interpolating shapes
-from the two transformed circles. For the condition r(ω) > 0, the
-pre-transformation values of r(ω) can be used. A scale transformation can
-flatten shapes to resemble lines. If a radial gradient is nested in the child
-sub-graph of a transformation that flattens the circles so that they are nearly
-lines, the centers may still be separated by some distance. In that case, a
-radial gradient would appear as a strip or a cone filled with a linear gradient.
+transformed and step 3 of the algorithm were performed by interpolating the
+shapes derived from the two transformed circles. For the condition r(ω) > 0, the
+pre-transformation values of r(ω) can be used.
+
+NOTE: A scale transformation can flatten shapes to resemble lines. If a radial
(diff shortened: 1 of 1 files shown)


OFF_AMD2_WD.md
```diff
@@ -468,11 +468,17 @@ extend mode. Figures 5.24 – 5.26 illustrate this for the different extend mode
 
 **Figure 5.26 Radial gradient with one circle contained within the other, reflect extend mode.**
 
-NOTE: A scale transformation (see 5.7.11.1.5) can flatten shapes to resemble
-lines. If a radial gradient is nested in the child sub-graph of a transformation
-that flattens the circles so that they are nearly lines, the centers may still
-be separated by some distance. In that case, a radial gradient would appear as a
-strip or a cone filled with a linear gradient.
+When a radial gradient is combined with a transformation (see 5.7.11.1.5), the
```
drott, line 471: As mentioned in our f2f meeting, with regards to this issue, I tend to think the math of the PaintTransformed and friends ought to be tight enough to describe exactly what happens with circles under transform and we don't need normative text using "shall" here. We can describe what follows from the normative text in PaintTransformed in the form of a note here, but if there's something unclear from how PaintTransformed and friends are working - then I believe we should fix it there, not write additional normative text here on how to specifically transform the circles. I am concerned we risk introducing inconsistencies otherwise. In other cases we used language like "One can think of it as..." - perhaps that would be useful here in describing how the radial gradient behaves under transforms?

OFF_AMD2_WD.md
```diff
@@ -467,17 +467,15 @@ extend mode. Figures 5.24 – 5.26 illustrate this for the different extend mode
...
 transformed and step 3 of the algorithm were performed by interpolating shapes
 from the two transformed circles. For the condition r(ω) > 0, the
-pre-transformation values of r(ω) can be used.
-
-NOTE: A scale transformation can flatten shapes to resemble lines. If a radial
-gradient is nested in the child sub-graph of a transformation that flattens the
-circles so that they are nearly lines, the centers may still be separated by
-some distance. In that case, a radial gradient would appear as a strip or a cone
-filled with a linear gradient.
+pre-transformation values of r(ω) can be used. A scale transformation can
```
drott, line 474: Optional: Put text from "A scale transformation..." into a second note? Seems to be a separate topic.

drott (approved):
Works for me this way, thanks.

OFF_AMD2_WD.md
PeterConstable: done

PR #223, opened 2021-01-25
For radial gradient, clarify if circles are identical

PeterConstable:
See #150—this case was already covered in the statement of the algorithm.


PeterConstable:
If we want a different behaviour, we'll need to revise step 1 in the algorithm.

PR #228, opened 2021-02-03
PaintSweepGradient: rename x -> centerX, y -> centerY

anthrotype:
`PaintRotate` and `PaintSkew` contain `centerX` and `centerY` for their respective center point, would be nice if also PaintSweepGradient used the same nomenclature for consistency.


PeterConstable:
Do you want to make the corresponding changes at line 239 in colr-gradients-spec.md?

anthrotype:
oh right, I'll do that

PeterConstable (approved):

anthrotype:
updated line 239 of colr-gradients-spec.md as well

PR #235, opened 2021-02-04
Rename PaintTransformed -> PaintTransform

anthrotype:
Sometimes we use "PaintTransform", other times "PaintTransformed".
I prefer to use "PaintTransform", without the final "ed". That's what I also used in the fontTools/ttx implementation.

Note that also the other transform-like paints are called PaintRotate, PaintSkew, PaintTranslate, not "PaintRotated", "PaintSkewed", etc.


PeterConstable (approved):
You've found all occurrences.

PR #249, opened 2021-02-05
Start revisions to add no-var formats

PeterConstable:
This is a start of the changes needed to add v1 formats without variation indices. This adds additional formats for gradients, but not yet for transforms.

I'll also need to look through the entire doc to see where else changes will be needed. And colr-gradients-spec.md will also need updating.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1507,14 +1507,16 @@ EXTEND_PAD by default.
 
 **5.7.11.2.5 Paint tables**
 
-Paint tables are used for COLR version 1 color glyph definitions. Twelve Paint
-table formats are defined (formats 1 to 12), each providing a different graphic
-capability for defining the composition for a color glyph. The graphic
-capability of each format and the manner in which they are combined to represent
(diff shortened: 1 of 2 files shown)


PeterConstable:
All of the additional formats needed for this design change have been added in 5.7.11.2.5. 

There are still ripple effects through the rest of the doc that would need to be addressed. Before working on that, though, we should close on going with this DC.

drott (approved):
LGTM.

I'm fine with landing this as is, but it would be good to have an update to https://github.com/googlefonts/colr-gradients-spec/#c-structures soon as well, either in this CL or a following one, as it makes for an easy quick reference when updating the implementation.

rsheeter:
FYI https://github.com/fonttools/fonttools/pull/2182/files is the nascent update to FontTools, including transforms

rsheeter (approved):

PeterConstable:
@rsheeter I've added changes to the C structures in colr-gradients-spec. Please review that. Then we can merge and I'll work on other ripple-effect changes in a separate PR.

rsheeter (approved):

PR #252, opened 2021-02-14
Additional changes in 5.7.11 COLR to incorporate non-variable formats

PeterConstable:
As I had to re-read all of 5.7.11, I made a few additional, very minor wording edits along the way.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -126,7 +126,7 @@ layer, and a red triangle with some transparency in the top layer.
 **Figure 5.8 Basic graphic capabilities of COLR version 0**
 
 The basic concepts also apply to color glyphs defined using the version 1
-formats: shapes are arranged in layers and have fills. But the additional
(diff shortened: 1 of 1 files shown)


OFF_AMD2_WD.md
```diff
@@ -1149,15 +1234,16 @@ root-to-leaf paths will be greater than or equal to the number of layers.
 The following are necessary for the graph to be well-formed and valid:
 
 * All subtable links shall satisfy the following criteria:
-  * Forward offsets are within the COLR table bounds.
+  * Forward offsets and non-NULL and are within the COLR table bounds.
```
drott, line 1237: I don't understand this addition. Did you mean "are non-NULL"? That's potentially incompatible with "COLR without CPAL", isn't it? I would prefer to defer adding this assertion.

drott (approved):
LGTM with some minor comments / optional suggestions.

anthrotype (approved):

PR #264, opened 2021-03-15
Remove restriction on varstore; compiler should have freedom to decide.

rsheeter:
Spec doesn't need to dictate choice of varstore, leave options open and let the compiler implementer decide. They'll have better data on the tradeoffs, may choose to emit multiple editions for different targets, etc.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2577,10 +2577,9 @@ The wordDeltaCount field contains a packed value that includes a flag and a
 
 The representation of delta values uses a mix of long types (“words”) and short
 types. If the LONG_WORDS flag is set, deltas are represented using a mix of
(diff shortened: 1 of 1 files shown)


rsheeter:
As @lorp notes you don't avoid overflow by disallowing this. Maybe you come up with a way to get better compression or higher performance due to alignment ... I don't think we should take the flexibility out at spec level, better to leave compilers to decide which store fits their needs..

drott (approved):
Agree with overflow considerations - LGTM.

PeterConstable:
@Lorp's observation is valid. It seems like a gap that this is not already covered in relation to instance value calculations in general. I've opened an [issue](https://github.com/MicrosoftDocs/typography-issues/issues/751) for the OT spec to fix this in OT 1.9.

PR #266, opened 2021-03-28
Improve description of CompositeMode enum

PeterConstable:
Revise the description per Vlad's comments


PeterConstable:
@vlevantovsky: please review

vlevantovsky (approved):
LGTM, with only minor caveat that in the AMD2 text (in the enumeration table) I will need to replace the references to the Compositing & Blending document with the bibliographic reference.

PeterConstable:
> I will need to replace the references to the Compositing & Blending document with the bibliographic reference.

I assumed something would be needed, and was leaving that detail for you.

rsheeter (approved):

PR #269, opened 2021-03-29
Add changes to OFF clause 7.1.7 for calculation of interpolated values

PeterConstable:


PeterConstable:
@vlevantovsky: please review

OFF_AMD2_WD.md
```diff
@@ -2398,6 +2399,21 @@ a gradient would use that level of quantization. Other factors from the
...
+_At the end of 7.1.7, insert the following text._
+
+In calculation of scalars (S, AS) and of interpolated values (scaledDelta,
+netAjustment, interpolatedValue), at least 16 fractional bits of precision
+should be maintained. Rounding should be done only when the final result is
+used. When scaled deltas are applied to a default value, the possibility of
+overflow exists. The integer bit-depth used in calculation shall be at least that
+of the data type of the item to which deltas are applied; for example, at least
+16 integer bits when applying scaled deltas to an FWORD value. Saturation
+arithmetic shall be used, with the final result clamped to the range of the
```
drott, line 2413: > Saturation
arithmetic shall be used, with the final result clamped to the range of the
data type of the item to which deltas are applied; for example, [-32768, 32,767]
for an FWORD value.

This phrasing would prevent or make less useful a higher internal representation / resolution / precision. Is that really intended?

PeterConstable: It doesn't limit precision; it only limits the numeric range.  I've added wording to clarify that fractional bits can be retained. 

It means, for example, that a contour point for an instance can't be moved to (36000, -40000), though the implementation could still represent the position of the contour point with fractional coordinates, e.g., (2345.678, 1357.246).

vlevantovsky (approved):
LGTM

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,29 @@ a gradient would use that level of quantization. Other factors from the
...
+to an FWORD value. Even if a final result is within range for the type of the
+default value, the possibility remains for overflow to occur at intermediate
+steps in calculation, and for the order in which deltas are added to affect the
+result. Sufficient integer bit-depth should be used to avoid any possible
+overflow at intermediate steps during calculation. An implementation-determined
+representation may be used for the final result (interpolatedValue), but
+saturation arithmetic must be used: values must not wrap from maximum to minimum
+values, or vice versa. The final result should be clamped to the range of the
+data type of the item to which deltas are applied; for example, [-32768, 32767]
+for an FWORD value.
```
drott, line 2432: After some internal discussion, some more feedback on this wording:

> The integer bit-depth used in calculation

The internal calculation may be in real space and should be allowed to be in real space.  - Integer bit-depth may be misleading? Is it necessary to specify that?

> the possibility remains for overflow to occur at intermediate steps in calculation

Suggest not use a new term intermediate steps in calculation.

> An implementation-determined
representation may be used for the final result (interpolatedValue), but
saturation arithmetic must be used: values must not wrap from maximum to minimum
values, or vice versa. The final result should be clamped to the range of the
data type of the item to which deltas are applied; for example, [-32768, 32767]
for an FWORD value.

I think we should allow the internal representation to be decoupled from the data type in the font format. FWORD may be used to transport and encode the font file, but internal computation of interpolated can operate in real space, on floats or doubles for example. With that, the implementation will not hit overflows, and would not need to clamp to the limits of the value that's used in the encoded font. Clamping to the constraints of the encoded file seems unnecessarily limiting. Implementations can address overflow by using floating point representations - which we probably to not want to mandate, but mandating clamping to external limits which the implementation can otherwise address is too strict too. Do we need to spec this?

PeterConstable: Re the first couple of comments, I'll reword.

Re recommendation on clamping the final result: it is stated as a recommendation, not a requirement. If the result is not clamped, then static fonts generated from a variation instance would display differently in such implementations than the variable font of which they are an instance. That wouldn't be good—hence the recommendation.

vlevantovsky: @PeterConstable 

> Re the first couple of comments, I'll reword.

Can you please propose a revision that we can review tomorrow during the break-out group discussion?
Thanks!

rsheeter:
New concern from f2f: in COLR and probably other places a variable value can be combined with other values (say a series of transforms) and a large value along the way might not be a problem (effectively add 40k, divide by 100, etc)

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,25 @@ a gradient would use that level of quantization. Other factors from the
...
 
+## Changes to OFF 7.1.7 Algorithm for interpolation of instance values
+
+_At the end of 7.1.7, insert the following text._
+
+An implementation-determined representation may be used during calculation and
+for the final result (interpolatedValue). In calculation of scalars (S, AS) and
+of interpolated values (scaledDelta, netAjustment, interpolatedValue), at least
+16 fractional bits of precision should be maintained. Rounding should be done
+only when the final result is used, and may retain greater fractional bit-depth
```
drott, line 2419: Optional:

```suggestion
16 fractional bits of precision should be maintained. If required for the internal representation 
rounding should be done only when the final result is used, and may retain greater fractional bit-depth
```

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,25 @@ a gradient would use that level of quantization. Other factors from the
...
+than that of the data type of the item to which deltas are applied. When scaled
```
drott, line 2420: Suggest to insert an empty line before "When scaled..."

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,25 @@ a gradient would use that level of quantization. Other factors from the
...
+deltas are applied to a default value, the possibility of overflow exists. The
```
drott, line 2421:
```suggestion
deltas are applied to a default value, the possibility of overflow exists: The
```

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,25 @@ a gradient would use that level of quantization. Other factors from the
...
+numeric range used in calculation must be at least that of the data type of the
+item to which deltas are applied; for example, at least [-32768, 32767] when
+applying scaled deltas to an FWORD value. Also, larger ranges should be allowed
+for during calculation to avoid any possible overflow at any point during
```
drott, line 2425: Repetitive "calculation"

```suggestion
applying scaled deltas to an FWORD value. Also, larger ranges should be allowed
for to avoid any possible overflow at any point during
```

OFF_AMD2_WD.md
```diff
@@ -2407,6 +2408,25 @@ a gradient would use that level of quantization. Other factors from the
...
+calculation, and to ensure that the order in which deltas are applied does not
+affect the final result. Saturation arithmetic shall be used: values shall not
+wrap from maximum to minimum values, or vice versa.
```
drott, line 2428: Optional: Saturation arithmetic on its own may be ambiguous if the internal representation is larger than the type in the font. Suggest to clarify. 

```suggestion
affect the final result. Saturation arithmetic shall be used respective to the internal representation: values shall not
wrap from maximum to minimum values, or vice versa.
```

drott (approved):
Thanks for the edits. Generally I don't have any major objections any longer as my requests to remove "intermediate steps" terminology and restricting the values to transport type are not in the text any more.

Some comments and suggestion, but looks good to me overall.

OFF_AMD2_WD.md
PeterConstable: Colon wouldn't be appropriate in that context: what follows isn't an elaboration or explanation.

OFF_AMD2_WD.md
PeterConstable: Accepted in principle: I've used slightly different wording.

PR #274, opened 2021-04-19
Detailed TOC for OFF_AMD2_WD.md

horasio:


horasio:
@googlebot I signed it!

horasio:
I understand. Still, since I'm currently implementing a COLRv1 renderer (in python for testing), I used my patched doc with clickable ToC a lot :-) I think it's useful while the spec is being refined and (heavily) used for testing purpose, as I do. (But it might be only me, in that case, merging the pull request is unnecessary.)

drott:
I find this useful as a first step. I am referring to this doc as well, and using it as the main reference, since we make changes during development of the spec here first. I agree it's going to be merged to OpenType which can serve as the reference, which has some of that structure, but while changes still happen, it's derived from the document here.

Issue #281, opened 2021-04-27
Color stop interval clarification?


justvanrossum:
I'm reading this over and over and I can't make sense of it:

> For example, a gradient color line could be defined with two color stops at 0.2 and 1.5. The gradient color line is positioned in the design grid by aligning stop offsets 0 and 1 to design grid positions, as defined for each gradient type, using an extrapolated color of stop offset 0 at one position and an interpolated color of stop offset 1 at the other position. Colors for offsets between 0.5 and 1.5 are interpolated. Colors for offsets above 1.5 and below 0.2 are defined and determined by the color line’s extend mode, described below.

(I think "0.5" should be "0.2".)

From the above, I truly don't understand what the intended behavior is for the example.

If I feed such values to Cairo, Skia, or write it in an SVG image (drawing a linear gradient), I observe the following among the three:
- Values outside of the [0, 1] interval appear to be clamped to [0, 1]
- Extending happens on the [0, 1] interval, regardless of the min/max stop offset values
- If the first stop offset is greater than zero, that color extends to the interval [0, firstStopOffset], *regardless of extend mode*
- Likewise, if the last stop offset is less than one, the last color extends to [lastStopOffset, 1], *regardless of extend mode*
- One exception: in Cairo, the above two behave differently in REPEAT mode (could that be a bug?), but is otherwise consistent with Skia and SVG for PAD and REFLECT.

What is the intention of the COLRv1 stop offsets when the total interval is not [0, 1]? Should it match the popular 2D platforms?

behdad:
The text you quoted is indeed bad.

The way we designed is, it is more generic and useful than the graphics library do, but is convertible to those. This is how it was intended:

Take the min/max stop positions and extend both ways towards infinity according to the extend mode. For PAD, the first and last are extended to -/+ inifinity. For REPEAT, the region between min/max is perpetually repeated in both directions. For REFLECT, the invert of the internal is appended on one end, and the resulting larger interval repeated in both directions.

So now you have a color-line that is well-defined for any input scalar number. Position the color-line on the graphics primitive such that the gradient's two points refer to the positions 0.0 and 1.0 on the color line. The colorline now specifies the color of the entire to infinity.

justvanrossum:
Thanks, it makes perfect sense now! So we pretty much refit the stop offsets from [min,max] to [0,1].

justvanrossum:
[...]

PR #292, opened 2021-06-02
rev wording describing color lines

PeterConstable:
Fix for #281

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -236,33 +236,37 @@ the successive color stops, to the last color stop. Between numerically-adjacent
 color stops, color values are linearly interpolated. See _Interpolation of
 Colors_ in 5.7.12 for requirements on how colors are interpolated.
 
+While the color gradation is specified over a defined interval, the color line
+continues indefinitely outside that interval in both directions. The color
+patterns outside the defined interval are determined by the color line’s *extend
+mode*, described below. In this way, colors are defined for all stop offset
+values, from negative infinity to positive infinity.
+
 For example, a gradient color line could be defined with two color stops at 0.2
-and 1.5. The gradient color line is positioned in the design grid by aligning
-stop offsets 0 and 1 to design grid positions, as defined for each gradient
-type, using an extrapolated color of stop offset 0 at one position and an
-interpolated color of stop offset 1 at the other position. Colors for offsets
-between 0.5 and 1.5 are interpolated. Colors for offsets above 1.5 and below 0.2
-are defined and determined by the color line’s *extend mode*, described below.
+and 1.5. Colors for offsets between 0.2 and 1.5 are interpolated. Colors for
+offsets above 1.5 and below 0.2 are determined by the color line’s extend mode.
+The gradient color line is positioned in the design grid by aligning stop
+offsets 0 and 1 to design grid positions. In this example, the color at stop
+offset 1 will be derived by interpolation between the two color stops, and the
+color at stop offset 0 will be derived based on the extend mode.
 
 If there are multiple color stops defined for the same stop offset, the first
 one is used for computing color values on the color line below that stop offset,
 and the last one is used for computing color values at or above that stop
 offset. All other color stops for that stop offset are ignored.
 
-While the color gradation is specified over a defined interval, the color line
-continues indefinitely outside that interval in both directions. The color
-pattern outside the defined interval is repeated according to the color line’s
-extend mode. Three extend modes are supported:
+The color patterns outside the defined interval are determined by the color
+line’s extend mode. Three extend modes are supported:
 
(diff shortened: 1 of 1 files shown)


PeterConstable:
@justvanrossum Please review

OFF_AMD2_WD.md
```diff
@@ -236,33 +236,37 @@ the successive color stops, to the last color stop. Between numerically-adjacent
 color stops, color values are linearly interpolated. See _Interpolation of
 Colors_ in 5.7.12 for requirements on how colors are interpolated.
 
+While the color gradation is specified over a defined interval, the color line
+continues indefinitely outside that interval in both directions. The color
+patterns outside the defined interval are determined by the color line’s *extend
+mode*, described below. In this way, colors are defined for all stop offset
+values, from negative infinity to positive infinity.
```
drott, line 243: Perhaps shortening both paragraphs (this and the previous), avoiding redundancy and talking about extrema as references helps to make clear that the minimum and maximum define the range and that 0 and 1 do not play a role? 

> Color gradation is defined over the interval from the first color stop, through
> the successive color stops, to the last color stop. Between numerically-adjacent
> color stops, color values are linearly interpolated. See _Interpolation of
> Colors_ in 5.7.12 for requirements on how colors are interpolated.

=>

Color gradation is defined over the interval from the minimum color stop, through
the successive color stops, to the maximum color stop. Between numerically-adjacent
color stops, color values are linearly interpolated. See _Interpolation of
Colors_ in 5.7.12 for requirements on how colors are interpolated.

Then shorter

```suggestion
The color
values for positions outside the defined interval from minimum to maximum are determined by the color line’s *extend
mode*, described below. In this way, colors are defined for all stop offset values, from negative infinity to positive infinity. 
```

OFF_AMD2_WD.md
```diff
@@ -236,33 +236,37 @@ the successive color stops, to the last color stop. Between numerically-adjacent
...
-type, using an extrapolated color of stop offset 0 at one position and an
-interpolated color of stop offset 1 at the other position. Colors for offsets
-between 0.5 and 1.5 are interpolated. Colors for offsets above 1.5 and below 0.2
-are defined and determined by the color line’s *extend mode*, described below.
+and 1.5. Colors for offsets between 0.2 and 1.5 are interpolated. Colors for
+offsets above 1.5 and below 0.2 are determined by the color line’s extend mode.
+The gradient color line is positioned in the design grid by aligning stop
+offsets 0 and 1 to design grid positions. In this example, the color at stop
+offset 1 will be derived by interpolation between the two color stops, and the
+color at stop offset 0 will be derived based on the extend mode.
```
drott, line 251: Suggest to strip "In this example, the color at stop
offset 1 will be derived by interpolation between the two color stops, and the
color at stop offset 0 will be derived based on the extend mode." as color at stop offset 0 is derived by extend mode, but also by interpolation between _repeated_ color stops.

rsheeter (approved):

OFF_AMD2_WD.md
```diff
@@ -231,38 +231,39 @@ only be specified within the range [-2, 2). See 5.7.11.2.4 for format details.)
...
-and 1.5. The gradient color line is positioned in the design grid by aligning
-stop offsets 0 and 1 to design grid positions, as defined for each gradient
-type, using an extrapolated color of stop offset 0 at one position and an
-interpolated color of stop offset 1 at the other position. Colors for offsets
-between 0.5 and 1.5 are interpolated. Colors for offsets above 1.5 and below 0.2
-are defined and determined by the color line’s *extend mode*, described below.
+and 1.5. Colors for offsets between 0.2 and 1.5 are interpolated. Colors for
+offsets above 1.5 and below 0.2 are determined by the color line’s extend mode.
+The gradient color line is positioned in the design grid by aligning stop
+offsets 0 and 1 to design grid positions.
```
justvanrossum, line 248: "design grid positions" mean "point 0" and "point 1" in the case of a linear gradient, right? For sweep gradients it's about angles, and the extend mode works differently if I understand correctly. For radial gradients, it's about the center points *and* the radii, correct?

In other words: The phrase "design grid positions" confused me because it's not explicit what it means for linear gradients, and is more unclear what it means for radial gradients, and is most confusing for sweep gradients.

rsheeter: I believe we clarify this per-type, would it suffice to delete "The gradient color line is positioned in the design grid by aligning stopoffsets 0 and 1 to design grid positions." entirely?

PeterConstable:
@justvanrossum Sorry, I didn't see your comment when I merged. 

I think better wording would be, "...by aligning stop offsets 0 and 1 to the design grid as defined for the given gradient type."

PeterConstable:
Re "design grid positions", see #296

PR #294, opened 2021-06-03
Remove virtual gids

### LINKED ISSUE #271: Remove virtual glyph-id talk
The hijacking of private (aka virtual) glyph ids by this table is something new to me. I oppose. And will not implement in HarfBuzz. I strongly advised that this nonsense be removed.

rsheeter:
Fixes #271. Rationale:

1. Allowing virtual gids isn't _useful_ for anything, notably it won't allow us to save gids
1. Complicates implementations; being able to use gid as globally unique is useful
1. ...I swear there was another reason but it's escaping me

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1435,6 +1435,9 @@ significantly different, however—see 5.7.11.1.
 | uint16 | glyphID | Glyph ID of the base glyph. |
 | Offset32 | paintOffset | Offset to a Paint table. |
 
+The glyphID value shall be less than the numGlyphs value in the &#39;maxp&#39;
+table (5.2.6).
+
 The records in the baseGlyphV1Records array shall be sorted in increasing
 glyphID order. It is intended that a binary search can be used to find a
 matching BaseGlyphV1Record for a specific glyphID.
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -50,8 +50,7 @@ COLR table; and the term “base glyph” is used to refer to a glyph for which 
 color glyph is provided. Processing of the COLR table is done on glyph sequences
 after text layout processing is completed and prior to final presentation of
 glyphs. Typically, a base glyph is a glyph that may occur in a sequence that
-results from the text layout process. In some cases, a base glyph may be a
-virtual glyph defined within this table as a re-usable color composition.
+results from the text layout process.
 
 For example, the Unicode character U+1F600 is the grinning face emoji. Suppose
 in an emoji font the 'cmap' table maps U+1F600 to glyph ID 718. Assuming no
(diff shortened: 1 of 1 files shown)


rsheeter:
@behdad you had explained the rationale for virtual gids being harmful to me and I then lost it. Could I impose on you to sanity check?

PeterConstable:
>Allowing virtual gids isn't useful for anything, notably it won't allow us to save gids

Rather than to save gids (which isn't the purpose of PaintColrGlyph), I thought the idea was to allow for defining a re-usable component without necessarily having to have a corresponding glyf entry. 

>Complicates implementations

In GSUB processing, font developers discovered they could use virtual gids even though not explicitly sanctioned by the spec and they just work without any complication to implementations. So in what way is implementation made more complex?

>being able to use gid as globally unique is useful

In what sense would a virtual gid not be globally unique?

Help me understand these concerns.

OFF_AMD2_WD.md
```diff
@@ -1815,14 +1804,12 @@ definitions. See 5.7.11.1.7.3 for more information.
 | Type | Name | Description |
 |-|-|-|
 | uint8 | format | Set to 11. |
-| uint16 | glyphID | Virtual glyph ID for a BaseGlyphV1List base glyph. |
```
PeterConstable, line 1818: This certainly shouldn't say "virtual"

behdad:
Rod, for the sake of process can I please ask you to commit the revert *without* any technical justification?

Then if Peter wants, he can propose them in an issue, describing what problem he's trying to solve, and why this is a good solution. I'd be happy to engage in technical discussion there.

I'm not trying to be a dick. Just have to draw a line somewhere. If this kind of sneaking-things-in is not tolerated in eg. Android / Chrome, I don't see why it's justified here.

PeterConstable:
I don't object to this change. I just don't know what problem is being solved.

I'm not trying to "sneak" anything in. With the change as proposed here, what I anticipate will happen is that implementations don't validate gid < numGlyphs and then people discover they can use virtual gids in their fonts and they just work. We've seen this happen since many years ago in GSUB. If we don't mind, then why not sanction it up front to ensure that it works everywhere?

If there are good reasons why fonts should not do that, then we should state explicitly that gid must be < numGlyphs. It strikes me, though, that when the spec has had such constraints people have come asking why or saying that the constraint is an unnecessary hindrance.

rsheeter:
I'm not sure this was snuck in, probably the reviewer [which was probably me] didn't realize it was an issue.

> a re-usable component without necessarily having to have a corresponding glyf entry

We mostly reuse via duplication (offset to the same paint) and slices into the shared layer list; I don't believe the virtual gid helps us here.

> implementations don't validate gid < numGlyphs ... we should state explicitly that gid must be < numGlyphs

We advise that the BaseGlyphV1Record gid can be used to compute bbox so I think we do want the gid to be valid (per maxp).

BaseGlyph, Layer, and PaintGlyph all specifiy gid must be valid per maxp. Ihave added a similar note to BaseGlyphV1Record. PaintColrGlyf must refer to a valid base glyph so I think that should resolve that?

PeterConstable:
> We advise that the BaseGlyphV1Record gid can be used to compute bbox so I think we do want the gid to be valid (per maxp).

Good point.

PeterConstable (approved):
The point about needing a real gid for bbox is a strong technical reason why not to allow virtual gids. This change (with the second commit) appears to cover all the places that should be touched.

PR #296, opened 2021-06-03
clarify color line wording re stops 0, 1

PeterConstable:


PeterConstable:
Cf. #292
@justvanrossum

justvanrossum:
That still misses that "design grid" is confusing/weird for sweep angles. @rsheeter's suggestion to simply delete the sentence isn't all bad, although feels somewhat unsatisfactory. I do appreciate the reference to the gradient types.

OFF_AMD2_WD.md
```diff
@@ -244,7 +244,7 @@ For example, a gradient color line could be defined with two color stops at 0.2
 and 1.5. Colors for offsets between 0.2 and 1.5 are interpolated. Colors for
 offsets above 1.5 and below 0.2 are determined by the color line’s extend mode.
 The gradient color line is positioned in the design grid by aligning stop
-offsets 0 and 1 to design grid positions.
+offsets 0 and 1 to the design grid as defined for the given gradient type.
```
rsheeter, line 247: Just exploring alternatives:

The application of the gradient color line is definied by the gradient type.

justvanrossum: Good enough for me.

rsheeter: @justvanrossum would that work to both avoid "design grid" confusion and guide reader to the per-gradient specification of?

justvanrossum: Yes, I think so.

PeterConstable: If we don't mention offsets 0 and 1 at all, it's probably better to just delete that sentence entirely. The intent was to indicate by the example that the color line could be defined with stops from offset min to offset max != 0, 1, but that the application was still done using offsets 0, 1. So, if we don't mention 0, 1, then the sentence is simply repeating something stated earlier in the section.

rsheeter: The one thing we lose by just deleting is the direction to see gradient type for how to apply the color line. I think probably that's fine though.

rsheeter (approved):
Just delete works for me.

Issue #299, opened 2021-06-09
Remove "V1" from member names


PeterConstable:
Back in #14, there was discussion about removing "V1" from field names, and seems like there was some agreement with removing (see [comment](https://github.com/googlefonts/colr-gradients-spec/issues/14#issuecomment-672993399) and following comments). But that seems to have gotten lost.

In the COLR v1 table, should these field names be revised?
- baseGlyphV1ListOffset -> baseGlyphListOffset
- layersV1Offset -> layerListOffset

And should these struct names be revised?
- BaseGlyphV1List table -> BaseGlyphList table
- LayerV1List table -> LayerList table

rsheeter:
I believe this is now resolved, please reopen if I am mistaken

PR #302, opened 2021-06-14
Revise struct names to remove 'V1'

PeterConstable:
Proposal re #299


rsheeter:
In f2f @drott pointed out V1 will be increasingly weird as we add V2+ so +1

rsheeter (approved):

rsheeter:
Could we do the C++ structs too plz?

PeterConstable:
@rsheeter C++ struct names revised; please review.

rsheeter (approved):

Issue #298, opened 2021-06-09
Too many of your bytes are belong to us


rsheeter:
COLR is, if I may paraphrase Just lightly, provides a powerful structure but seems a bit airy, feels like it can be compacted more. #283 covers one specific possible path. Two more to explore, probably together:

1. Introduce a VarIdxMap (ala HVAR)
   * store a single VarId in the Paint graph and lookup VarId, VarId+1, etc in the map to get however many variation indices are needed.
   * VarIdxMap may achieve better packing than our current mandatory 4 byte per field VarWhatever types
1. Consider lower precision for specialized transforms
   * For example, to reuse a shape offset is rather fundamental, we immediately see *tons* of them when exploring variable COLR construction
       * We could amend PaintTranslate (or have a new Paint) that offers FWORD dx, dy and perhaps has the child Paint directly following (no Offset24). 5 bytes where today we spend 12.

Credit @behdad for both the initial ideas and this diagram to clarify them:

![image](https://user-images.githubusercontent.com/6466432/121291112-94183580-c89c-11eb-825e-e147ee1531a8.png)

I want to do some data gathering and if it looks promising I'll put up a PR for potential adjustments.

behdad:
I'm still not happy with this. We're still trying to use VarStore for what it wasn't designed for (storage of tuples of variations). But I hope this extra VarIdxMap facility can save some of the pressure.

Looking at the CJK font we got, the biggest saving will come from using an FWORD PaintTranslate, which would also shrink the variations stored in the VarStore, and then the VarIdxMap can save bytes if we adjust the VarStore optimizer; that is, write a VarIdxMap-aware optimizer. Right now we don't have that. I'm thinking that through.

justvanrossum:
Is the additional indirection via a VarIdxMap even necessary? This may be naïve, but I imagine a single "outer" index, and an incrementing sequence of "inner" indices.

behdad:
> Is the additional indirection via a VarIdxMap even necessary?

Without it, all the variations within the same paint have to use the same outer index, which limits optimization opportunity. That said...

> This may be naïve, but I imagine a single "outer" index, and an incrementing sequence of "inner" indices.

This is exactly what it will boil down to when the optional `VarIdxMap` is not present.

PeterConstable:
[...]

PR #304, opened 2021-06-28
Add low precision translation

rsheeter:
Based on investigation conducted based on discussion in #298.


rsheeter:
I think #307 does a better job

Issue #283, opened 2021-04-29
Contemplate reduction of indirection


rsheeter:
@behdad was wondering about a model where we reduce indirection; we may have considerable use of offsets that are unnecessary. Would like to count by paint type and ideally seek chains where the offset is pointless (the paint could just follow inline).

PeterConstable:
I assume size-savings (avoid an Offset32) isn't the concern. (I don't think we'd want to define a parallel pairs of offset-less formats.) We could specify that an offset of 0 means that the paint subtable immediately follows the current paint table.

Is this really a v1 issue?

behdad:
Wrong assumption.

PeterConstable:
We're up to 32 paint formats now (after adding PaintScale etc.). Formats 1 to 9 (solid, gradients) and 11 (PaintColrGlyph) don't have a fields for paint subtable offsets. The formats that do are:
* PaintGlyph (format 10)
* transforms (formats 12 to 31)
* PaintComposite (format 32) — has two offset fields.

We already doubled the transform formats to have var/no-var pairs. And we've doubled some further for center/no-center pairs. We'd be adding 20 more transform formats.

The gradient formats also have offsets to ColorLine subtables. Doubling those would be six more formats.

A potential concern is the precedent: how far do we go in future OT enhancements creating variants of formats for different combinations of optional fields?

rsheeter:
Size savings _may_ be a concern. I aspire to measure but I haven't had time to actually do it yet. Too many priorities lol.

I think we should let the data speak, if it saves 1% (of the overall font) we can forget this. If it saves a significant amount maybe it's worth thinking about how we might support no-offset variations. We could invest a format bit, we could say referenced paints *always* follow except when a special format (jump, PaintOffset as it were) is found, ... I'm sure there are other options.

Wrt number of formats I think it's important to remember that an implementation isn't obliged to define structures per format. One could very reasonably unpack every many transform formats into the same structure. One could unpack SomethingNotVariable and SomethingVariable into the same struct, etc. A compiler might have to care about all variants but that's fine, more flexibility to the compiler is a good thing.

rsheeter:
Not quite ready to close but so far my experiments are making this appear unappealing.

behdad:
I would have encouraged doing this for Translate at least *if* there was no further technical issue. But if we do it as is, the base of the offsets in subsequent Paint's become unclear. That's a pain.

PeterConstable:
[...]

PR #305, opened 2021-06-28
Make the 2x3 an offset

rsheeter:
Based on exploration done for #283, #298 gains from sharing structures tends to outweigh trimming offsets

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1847,7 +1847,7 @@ see 5.7.11.1.5.
 |-|-|-|
 | uint8 | format | Set to 12. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| Affine2x3 | transform | An Affine2x3 record (inline). |
+| Offset24 | transform | Offset to a Affine2x3. |
 
 *PaintVarTransform table (format 13):*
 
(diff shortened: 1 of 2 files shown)


OFF_AMD2_WD.md
```diff
@@ -1847,15 +1847,15 @@ see 5.7.11.1.5.
...
 *PaintVarTransform table (format 13):*
 
 | Type | Name | Description |
 |-|-|-|
 | uint8 | format | Set to 13. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| VarAffine2x3 | transform | A VarAffine2x3 record (inline). |
+| Offset24 | transform | Offset to a VarAffine2x3. |
 
 The affine transformation is defined by a 2×3 matrix, specified in an Affine2x3
```
drott, line 1860: Optional: I am not sure how we described it in other offset cases, but in this text and analogously we may want to explicitly mention that the actual Affine2x3 is pointed to buy an offset and not somehow "in" the Offset24 variable.

drott (approved):
This, together with the reductions in the other PR look like reasonable changes to me if we prefer to go for compact formats while the 2x3 transform is the high precision option.

OFF_AMD2_WD.md
rsheeter: I tried to match the paint one, which reads "Offset to a Paint subtable." I'll sanity check a few more to be sure they aren't just both inconsistent but I believe this is how we've described offsets throughout.

rsheeter: Looks like we use "offset to a" in both text and table descriptions

PR #307, opened 2021-06-28
Reduce precision for all transforms added for compactness

rsheeter:
Exploit the follow observations from test fonts:

1. The most common translation is by an integer that fits within int16 so change dx/dy to Int16
1. The most common rotation is by < 360° so change angles to F2Dot14 where 1.0 means 180° counter-clockwise
1. The most common scales are small, it's rare to want a massively larger/smaller version of something, so use F2Dot14 scales
1. The most common center for a transformation is an integer that fits within int16 so change centerX/Y to Int16

Note that:

1. In all cases full precision can be achieved by using the full 2x3 whose types are unchanged.
1. Variation can be used to access greater range, such as if you want an axis that causes many rotations.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1913,7 +1913,7 @@ provides a more compact representation when variation is not required. Format 15
 shall not be used in non-variable fonts or if the COLR table does not have an
 ItemVariationStore subtable.
 
-These tables used reduced precision for compactness. Where higher precision is
+These tables use reduced precision for compactness. Where higher precision is
 required use PaintTransform/PaintVarTransform.
 
 For general information regarding transformations in a color glyph definition, 
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1973,7 +1973,7 @@ is not required.
 Formats 17, 19, 21 and 23 shall not be used in non-variable fonts or if the COLR
 table does not have an ItemVariationStore subtable.
 
-These tables used reduced precision for compactness. Where higher precision is
+These tables use reduced precision for compactness. Where higher precision is
 required use PaintTransform/PaintVarTransform.
 
 For general information regarding transformations in a color glyph definition, 
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1925,8 +1925,8 @@ see 5.7.11.1.5.
 |-|-|-|
 | uint8 | format | Set to 14. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| Int16 | dx | Translation in x direction. |
-| Int16 | dy | Translation in y direction. |
+| FWORD | dx | Translation in x direction. |
+| FWORD | dy | Translation in y direction. |
 
 *PaintVarTranslate table (format 15):*
 
@@ -1934,8 +1934,8 @@ see 5.7.11.1.5.
 |-|-|-|
 | uint8 | format | Set to 15. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| VarInt16 | dx | Translation in x direction. |
-| VarInt16 | dy | Translation in y direction. |
+| VarFWORD | dx | Translation in x direction. |
+| VarFWORD | dy | Translation in y direction. |
 
 NOTE: Pure translation can also be represented using the PaintTransform or
 PaintVarTransform table by setting _xx_ = 1, _yy_ = 1, _xy_ and _yx_ = 0, and
@@ -2005,8 +2005,8 @@ see 5.7.11.1.5.
 | Offset24 | paintOffset | Offset to a Paint subtable. |
 | F2Dot14 | scaleX | Scale factor in x direction. |
 | F2Dot14 | scaleY | Scale factor in y direction. |
-| Int16 | centerX | x coordinate for the center of scaling. |
-| Int16 | centerY | y coordinate for the center of scaling. |
+| Fixed | centerX | x coordinate for the center of scaling. |
+| Fixed | centerY | y coordinate for the center of scaling. |
 
 *PaintVarScaleAroundCenter table (format 19):*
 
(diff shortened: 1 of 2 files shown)


Lorp:
Careful about rotation centres… wanting 0.5 unit precision seems reasonable for the simple use case of rotating about the glyph centre.

OFF_AMD2_WD.md
```diff
@@ -1970,6 +1973,9 @@ is not required.
 Formats 17, 19, 21 and 23 shall not be used in non-variable fonts or if the COLR
 table does not have an ItemVariationStore subtable.
 
+These tables used reduced precision for compactness. Where higher precision is
```
drott, line 1976:
```suggestion
These tables use reduced precision for compactness. Where higher precision is
```

OFF_AMD2_WD.md
```diff
@@ -1979,75 +1985,75 @@ see 5.7.11.1.5.
 |-|-|-|
 | uint8 | format | Set to 16. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| Fixed | scaleX | Scale factor in x direction. |
-| Fixed | scaleY | Scale factor in y direction. |
+| F2Dot14 | scaleX | Scale factor in x direction. |
+| F2Dot14 | scaleY | Scale factor in y direction. |
```
drott, line 1989: Can you share some additional histogram data from the experimentation about range of values and distribution, reducing this scale paint to -1.99996948242 to 1.99996948242 (if I am not mistaken about the F2Dot14 range) seems a strong limitation - more data would help on common ranges may help. Even though we're coming from a particular set of source SVGs. Other font productions may work differently but most of it can be handled by the compiler by prescaling, I assume.

OFF_AMD2_WD.md
```diff
@@ -2073,6 +2080,9 @@ when variation is not required.
 Formats 25 and 27 shall not be used in non-variable fonts or if the COLR table
 does not have an ItemVariationStore subtable.
 
+These tables used reduced precision for compactness. Where higher precision is
```
drott, line 2083:
```suggestion
These tables use reduced precision for compactness. Where higher precision is
```

OFF_AMD2_WD.md
```diff
@@ -2082,35 +2092,35 @@ see 5.7.11.1.5.
 |-|-|-|
 | uint8 | format | Set to 24. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| Fixed | angle | Rotation angle, in counter-clockwise degrees. |
+| F2Dot14 | angle | Rotation angle, 180° in counter-clockwise degrees per 1.0 of value. |
```
drott, line 2095: This seems good to me from a space change, in particular because anything that's wildly over 360° in `Fixed` space does not make sense and wastes spaces. But I am wondering if we introduce problems if we can't can't reach 360° with F2Dot14 with a positive or negative value? Would it?

OFF_AMD2_WD.md
```diff
@@ -2082,35 +2092,35 @@ see 5.7.11.1.5.
...
 
 | Type | Name | Description |
 |-|-|-|
 | uint8 | format | Set to 26. |
 | Offset24 | paintOffset | Offset to a Paint subtable. |
-| Fixed | angle | Rotation angle, in counter-clockwise degrees. |
-| Fixed | centerX | x coordinate for the center of rotation. |
-| Fixed | centerY | y coordinate for the center of rotation. |
+| F2Dot14 | angle | Rotation angle, 180° in counter-clockwise degrees per 1.0 of value. |
+| Int16 | centerX | x coordinate for the center of rotation. |
```
drott, line 2112: @Lorp can you elaborate on your comment? 

I think `Int16` here seems reasonable to use as a center coordinate while still be able to say "rotate around the center" of the glyph. Or was your concern related to the `F2Dot14` for the rotation angle?

drott (commented):
Generally seems like a good change to me, thanks for incorporating that and studying the findings. Some more data on `PaintScale` and friends may be useful.

OFF_AMD2_WD.md
Lorp: The point is that the centre of a glyph is non-integral whenever that dimension measures an odd number. For example, consider the glyph that is a square with diagonal (1,1) to (100,100), with centre (50.5,50.5). In general, rotation centres have two integral coordinates only 25% of the time. Rotating accurately about the glyph centre seems to me to be a desirable operation. How about a 2x factor for this `Int16` ? :)

OFF_AMD2_WD.md
rsheeter: 0x8000 is -2.0 or -360° so I believe we have access to a complete rotation. For a static it doesn't matter which way you rotated and for a variation you have the added variation value to spin whatever way you like.

OFF_AMD2_WD.md
rsheeter: F2Dot14 is [-2.0, 1.999939] if https://docs.microsoft.com/en-us/typography/opentype/spec/otff is to be believed.

Data is skewed to one particular font but as that font is very deliberately heavily componentized I think that's not too big of a problem. In that case 99+% of scales are within range. This matches intuitive expectation: scaling something to half size or 1.5x in text would be much more common than wanting a copy that is 10x larger. Further, if you do want two components where one is 10x the size of the other you have two good options:

1. Write the outline for the big one and scale it down (0.1x the big one instead of 10x the small one); you get greater precision on the outline this way
1. Use the full 2x3

Using the variation version to increase range is also possible but not really recommended I wouldn't think.

drott: Ack, thanks, makes sense.

OFF_AMD2_WD.md
rsheeter: I think I'm inclined to agree. Although it did not pop out in testdata the ability to spin around point is *whips off glasses* the point :D Let me think about it just a little longer but I'm leaning toward making the center of rotation only Fixed again.

behdad:
I find the Fixed for rotation center overkill. But whatever :).

OFF_AMD2_WD.md
```diff
@@ -1913,6 +1913,9 @@ provides a more compact representation when variation is not required. Format 15
 shall not be used in non-variable fonts or if the COLR table does not have an
 ItemVariationStore subtable.
 
+These tables used reduced precision for compactness. Where higher precision is
```
drott, line 1916:
```suggestion
These tables use reduced precision for compactness. Where higher precision is
```

OFF_AMD2_WD.md
drott: Ack re angles.

Remaining nit, line length / wrap if possible, here and below.

drott (approved):
Works for me, LGTM.

rsheeter:
> I find the Fixed for rotation center overkill

@behdad could you expand? Would you contend FWORD sufficient? - that is the one thing I was really waffling on.

rsheeter:
I'm migrating the FWORD center discussion to #308 so I can merge the other changes.

anthrotype:
The angle of PaintSweepGradients is still a `Fixed`, should it also be a F2Dot14?

PR #310, opened 2021-06-30
All angles in F2Dot14

rsheeter:
Make all the angle fields consistent.

Smallest rotation in degrees (ty @anthrotype for pointing this out)

```
>>> 1/(1<<14)*180
0.010986328125
```

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1768,8 +1768,8 @@ formats, see 5.7.11.2.4. For background information on the color line, see
 | Offset24 | colorLineOffset | Offset to ColorLine table. |
 | FWORD | centerX | Center x coordinate. |
 | FWORD | centerY | Center y coordinate. |
-| F2Dot14 | startAngle | Start of the angular range of the gradient. 180° in counter-clockwise degrees per 1.0 of value. |
-| F2Dot14 | endAngle | End of the angular range of the gradient. 180° in counter-clockwise degrees per 1.0 of value. |
+| F2DOT14 | startAngle | Start of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. |
+| F2DOT14 | endAngle | End of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. |
 
 *PaintVarSweepGradient table (format 9):*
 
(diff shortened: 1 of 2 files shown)


anthrotype:
We still have this paragraph in PaintSweepGradient

> Start and end angle values can be outside the range [0, 360), but are interpreted as values within that range by applying a modulus operation. For example, an angle -60° is treated the same as 300°; an angle 480° is treated the same as 120°. As a consequence, the [0, 1] range of the color line covers at most one full rotation around the center, never more.

But since the start/end angles are now encoded as F2Dot14 ranging from [-2..2) and interpreted as fractions of 180 degrees, they can never specify angles less than -360 or greater than 359.989013671875 (*), so that paragraph is outdated and might need rewording.

***

(*) A F2Dot14 is packed as a signed 16-bit integer, so it can express -32768 <= value <= 32767. Converting fixed to floats using 14 bits for the fractional part (using fontTools' fixedTools module) you get this for the min/max

fixedToFloat(-32768, precisionBits=14) == -2.0
fixedToFloat(32767, 14) == 1.99993896484375

Multiply by 180 to obtain degrees and we get respectively -360 and +359.989013671875 as minimum and maximum angle (in degrees) that can be encoded with this fraction-of-half-circle angle notation.

rsheeter:
it's still true you can obtain values outside that range under variation; amended to specifically mention that

OFF_AMD2_WD.md
```diff
@@ -622,8 +622,8 @@ in figure 5.30:
 
 **Figure 5.30 A sweep gradient with start angle of 30° and an end angle of 150°**
 
-Start and end angle values can be outside the range [0, 360), but are
-interpreted as values within that range by applying a modulus operation. For
+Start and end angle values can be outside the range [0, 360) when variations are applied.
```
PeterConstable: As @anthrotype pointed out, the F2DOT14 representation for angles allows a range [-2, 2), which corresponds to [-360º, 360º). So even apart from variations angles can be outside [0, 360).

rsheeter: Ah good point, so we don't actually need to change anything there

rsheeter: ...and it look slike you fixed it, ty

PeterConstable (approved):

PR #315, opened 2021-07-02
fix typo 'with' -> 'within'

PeterConstable:


vlevantovsky (commented):
There is another typo that needs fixing - in the Radial Gradients sub-clause, in second paragraph the word "with" is duplicated: "... but adapted with with alternate color line extend modes"

PeterConstable:
Thanks; fixed in #317

vlevantovsky:
Thanks!

PR #306, opened 2021-06-28
Introduce a varidxmap

rsheeter:
To facilitate better packing, introduce VarIdBase & VarIdxMap, ref #298.

PR #316, opened 2021-07-02
use DeltaSetIndexMap instead of VarFWord, etc. structs

PeterConstable:
With the fifth commit, this is **NOW COMPLETE**.

This will supersede PR #306.

In the section on COLR and variations (5.7.11.4), some details regarding potential edge cases needed to be covered.

>If the COLR table does not contain an Item Variation Store subtable, the _varIndexBase_
field of variable tables or records shall be ignored by applications, and should be set to
zero.

This maintains what was previously there but re-cast into the new model.

>If the index for a variable item is greater than or equal to the number of entries in the mapping array, the last mapping array entry shall be used.

This follows what is already done in [HVAR](https://docs.microsoft.com/en-us/typography/opentype/spec/hvar).

>The sequence of indices derived from a _varIndexBase_ value do not wrap on overflow and shall not exceed 0xFFFFFFFF. A _varIndexBase_ value of 0XFFFFFFFF is assigned a special meaning indicating that the variable fields in the given table or record do not have variation data.
> ...
>If the COLR table contains an Item Variation Store but does not contain a mapping table, then an implicit identity mapping is used: the sequence of values beginning with _varIndexBase_ are treated directly as delta-set indices with 16-bit sub-fields for _outer_ (high word) and _inner_ (low word) index values (see 7.2.3.3).

These are new and proposed.

In HVAR, if there is an IVS but not a mapping table, then the indices into the IVS are implicit, using 0 as outer index and the gid as the inner index. That won't work for COLR. So, what _could_ work? If _varIndexBase_ is 32-bit, then that means it could be directly as an IVS outer/inner index. That's what I'm proposing here.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2499,7 +2499,7 @@ For example, the VarAffine2x3 table (5.7.11.2.5.8) has eight variable fields fol
 
 If the index for a variable item is greater than or equal to the number of entries in the mapping array, the last mapping array entry shall be used.
 
-The sequence of indices derived from a _varIndexBase_ value do not wrap on overflow and shall not exceed 0xFFFFFFFF. A _varIndexBase_ value of 0XFFFFFFFF is assigned a special meaning indicating that the variable fields in the given table or record do not have variation data.
+The sequence of indices derived from a _varIndexBase_ value do not wrap on overflow and shall not exceed 0xFFFFFFFF. A _varIndexBase_ value of 0xFFFFFFFF is assigned a special meaning indicating that the variable fields in the given table or record do not have variation data.
 
 Similarly, a delta-set index mapping entry with values 0xFFFF/0xFFFF can be used to indicate that an item has no variation data. (See 7.2.3.2.)
 
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1296,10 +1296,13 @@ that may be rendered as a non-color glyph instead.
 **5.7.11.2 COLR table formats**
 
 Various table and record formats are defined for COLR version 0 and version 1.
-Several values contained within the version 1 formats are variable. These use
-various record formats that combine a basic data type with a variation delta-set
-index: VarFWord, VarUFWord, VarF2Dot14, and VarFixed. These are described in
-7.2.3.1.
+
+Several values contained within the version 1 formats are variable. For items
+that vary in a variable font, the variation data is contained in an
+ItemVariationStore table (7.2.3). To associate each variable item with the
+corresponding variation data, a DeltaSetIndexMap table is used, as in the HVAR
+table (7.3.5). Within a given table that has variable items, a base/sequence
+scheme is used to index into the mapping data. See 5.7.11.4 for details.
 
 All table offsets are from the start of the parent table in which the offset is
 given, unless otherwise indicated.
@@ -1337,6 +1340,7 @@ the COLR table require glyph ID 1 to be the .null glyph.
 | uint16 | numLayerRecords | Number of Layer records; may be 0 in a version 1 table. |
 | Offset32 | baseGlyphListOffset | Offset to BaseGlyphList table. |
 | Offset32 | layerListOffset | Offset to LayerList table (may be NULL). |
+| Offset32 | varIndexMapOffset | Offset to DeltaSetIndexMap table (may be NULL). |
 | Offset32 | itemVariationStoreOffset | Offset to ItemVariationStore (may be NULL). |
 
 The BaseGlyphList and its subtables are only used in COLR version 1.
@@ -1349,6 +1353,12 @@ NULL.
 The ItemVariationStore is used in conjunction with a BaseGlyphList and its
 subtables, but only in variable fonts. If it is not used, set
 itemVariationStoreOffset to NULL.
+
+The DeltaSetIndexMap format is described in 7.3.5.2. A DeltaSetIndexMap is used
+in conjunction with the ItemVariationStore in a variable font. The
+DeltaSetIndexMap is optional: if an ItemVariationStore is present but a 
+DeltaSetIndexMap is not included, then an implicit mapping is used. See
+5.7.11.4 for details.
 
 **5.7.11.2.1.3 Mixing version 0 and version 1 formats**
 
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM

colr-gradients-spec.md
```diff
@@ -233,12 +229,13 @@ struct Affine2x3
...
-  VarFixed yy;
-  VarFixed dx;
-  VarFixed dy;
+  Fixed xx;  // VarIdx varBase + 0
+  Fixed yx;  // VarIdx varBase + 1
+  Fixed xy;  // VarIdx varBase + 2
+  Fixed yy;  // VarIdx varBase + 3
+  Fixed dx;  // VarIdx varBase + 4
+  Fixed dy;  // VarIdx varBase + 5
+  VarIdBase varBase;
```
anthrotype, line 238: the typedef is `VarIdxBase`, not `VarIdBase`. The latter typo appears multiple times here and below

colr-gradients-spec.md
```diff
@@ -105,7 +105,11 @@ struct ArrayOf
...
 
-typedef uint32 VarIdx;
+// Index of the first variation record in the variation index map or
+// variation store if no variation index map is present.
+// A record with N variable fields finds them at VarIdBase+0, ...,+N-1
+// Use 0xFFFFFFFF to indicate no variation.
+typedef uint32 VarIdxBase;
 
 template <typename T>
 struct Variable
```
anthrotype, line 115: this `Variable` template is no longer used, can be removed I guess

OFF_AMD2_WD.md
```diff
@@ -1548,8 +1563,9 @@ stop offset position or of alpha, and one that does not.
 
 | Type | Name | Description |
 |-|-|-|
-| VarF2Dot14 | stopOffset | Position on a color line; variable. |
-| VarColorIndex | color | |
+| F2DOT14 | stopOffset | Position on a color line. For variation, use varBaseIndex + 0. |
+| ColorIndex | color | For variation, use varBaseIndex + 1. |
+| uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
```
anthrotype, line 1568: the struct field name is called `varIndexBase` here, and `varBase` in the C++ header, I suggest we use consistent naming in both

OFF_AMD2_WD.md
```diff
@@ -1337,6 +1340,7 @@ the COLR table require glyph ID 1 to be the .null glyph.
 | uint16 | numLayerRecords | Number of Layer records; may be 0 in a version 1 table. |
 | Offset32 | baseGlyphListOffset | Offset to BaseGlyphList table. |
 | Offset32 | layerListOffset | Offset to LayerList table (may be NULL). |
+| Offset32 | varIndexMapOffset | Offset to DeltaSetIndexMap table (may be NULL). |
```
anthrotype, line 1343: the same table is called sometimes `DeltaSetIndexMap` other times `VarIdxMap` (in the c++ header) or `varIndexMap` (as in the name of the `varIndexMapOffset` field above). FontTools otData.py calls it `VarIdxMap`, whereas the current OT spec calls it `DeltaSetIndexMap`.
Using multiple names for the same thing may lead to confusions

PeterConstable: I don't have any strong opinion on how it's referred to in the C++ header or in otData.py, though I agree consistency is better. But it's already called DeltaSetIndexMap in the OT spec, and so that's what should be used in the COLR spec.

PeterConstable: At line 551 in the C++ header, "VarIdxMap" is used; but I don't know if that's because "VarIdxMap" was used in some C++ implementation elsewhere. So, I'll leave that to someone else to sort out.

OFF_AMD2_WD.md
```diff
@@ -1548,8 +1563,9 @@ stop offset position or of alpha, and one that does not.
 
 | Type | Name | Description |
 |-|-|-|
-| VarF2Dot14 | stopOffset | Position on a color line; variable. |
-| VarColorIndex | color | |
+| F2DOT14 | stopOffset | Position on a color line. For variation, use varBaseIndex + 0. |
```
vlevantovsky, line 1566: In "description" fields, varBaseIndex should be varIndexBase.

OFF_AMD2_WD.md
```diff
@@ -1548,8 +1563,9 @@ stop offset position or of alpha, and one that does not.
...
+| ColorIndex | color | For variation, use varBaseIndex + 1. |
```
anthrotype, line 1567: shoudn't `VarColorStop` contain a `VarColorIndex` instead of a simple `ColorIndex`? The variability is embedded in the inner `VarColorIndex` (which in turn defines its own varIndexBase, distict from VarColorStop.varIndexBase)

PeterConstable: I followed what Rod had drafted in #306, and I think I see why he did this. There's a bit of a quirk regarding ColorIndex vs. VarColorIndex and the contexts in which they are used, though, and we should probably discuss this further, but _after_ we complete this change (for what needs to go into the CDAM text—we can change in ballot comments if needed). 

In VarColorStop, you could use VarColorIndex, but then VarColorStop.varIndexBase would only be used for VarColorStop.stopOffset. You'd end up with two occurrences of varIndexBase (the second embedded: VarColorStop.color.varIndexBase) each of them being used for exactly one member.  By having VarColorStop use ColorIndex, you can have a single varIndexBase being used for both VarColorStop.stopOffset and VarColorStop.color.alpha. 

(I had meant to clarify that in the description for VarColorStop.color but forgot. I also overlooked the mention at line 319.)

It might also be possible to dispense with VarColorIndex entirely: In PaintVarSolid, you could put varIndexBase into PaintVarSolid and have it affect PaintVarSolid.color.alpha in just the same way.

But having a member affect the interpretation of a peer's embedded member (whether in PaintVarSolid or here, in VarColorStop) maybe isn't best.

I've opened #318 as a separate issue.

vlevantovsky: > There's a bit of a quirk regarding ColorIndex vs. VarColorIndex and the contexts in which they are used, though, and we should probably discuss this further, but after we complete this change (for what needs to go into the CDAM text—we can change in ballot comments if needed).

We can discuss and introduce any additional changes during the WG meeting week, we don't need to wait to submit it as a ballot comment. If this is ready by next Monday, we can add this as an additional agenda item for Font break-out session discussion and implement whatever changes we will have agreed on at that time.

PeterConstable: Sure; that makes sense. But I'd like to complete this PR and then address this separately in discussion of #318.

Issue #318, opened 2021-07-07
use ColorIndex or VarColorIndex in VarColorStop?


PeterConstable:
In VarColorStop, a varIndexBase member is needed for the stopOffset. Should that base also be used for the embedded alpha? That is, should VarColorStop have ColorStop as a member, with VarColorStop.varIndexBase also applied to VarColorStop.color.alpha?

Or, should VarColorStop have a VarColorStop member, with VarColorStop.varIndexBase used only for VarColorStop.stopOffset, and separately have VarColorStop.color.varIndexBase used for VarColorStop.color.alpha? (So, two separate varIndexBase members in the two structs, each used for exactly one member.)

(Similar question for PaintVarSolid: if it had a varIndexBase member that applied to its color.alpha, then perhaps VarColorIndex wouldn't be needed at all.)

rsheeter:
Per f2f I'm tempted to suggest dropping ColorIndex completely, just list the fields.

rsheeter:
Fixed by #320

PR #320, opened 2021-07-13
Drop ColorIndex, flatten color stop & paint solid structures

PeterConstable:
Changes for #318 per F2F discussion.

Changes at lines 260 – 269 and 1550 – 1553 only correct to the desired line wrap in .md source—no substantive changes.


rsheeter:
LGTM, maybe wait a day or so in hopes @anthrotype or @drott might have a chance to glance over as well?

rsheeter (approved):

Issue #289, opened 2021-05-24
With variability, pre-computing a bounding box is non-trivial


justvanrossum:
I'm not sure what I'm asking here, but I'd like to raise awareness that pre-calculating a (variable) bounding box for a variable color glyph (to put in glyf+gvar or CFF2, as required by the spec) is non-trivial.

Not only that, but the tightest possible bounding box may not be expressible without higher order interpolation.

But even calculating a non-tight bounding box that is guaranteed to always be large enough is non-trivial, for example in the case of a rotating element using PaintRotate.

The burden for authoring tools to get this right is pretty heavy, if you ask me.

Am I overthinking this? Does anyone have ideas of how to even approach this in a general way?

rsheeter:
I believe you are right that it's difficult and indeed we don't even do it right for static in our current implementation. However, I'm not (yet) convinced it's infeasible. My instinct is to say "leave it to the compiler" and the compiler can evolve independent of the spec.

https://github.com/googlefonts/nanoemoji/issues/238
https://github.com/googlefonts/nanoemoji/issues/239

PeterConstable:
Not sure there's any alternative.

rsheeter:
I think the right thing to do is close here - no spec change anticipated - and pursue further in the context of the public compiler (tickets above). Please reopen if you disagree.

justvanrossum:
I agree, sorry for not closing sooner.

drott:
Commenting here as well after Rod closed https://github.com/googlefonts/nanoemoji/issues/238#issuecomment-912049713 The new changes in the spec for explicit ClipBox or not specifying a ClipBox provide a solution for it. Precomputing and adding a ClipBox can be a useful optimisation. For variable scenarios, not specifying a ClipBox is an option in which case the rasteriser traverses the graph to determine a bounding box.

PR #321, opened 2021-07-14
Give COLR an explicit bbox to call it's own

### LINKED ISSUE #251: Consider allowing indirection to retrieve bounding box from different glyf glyph
In a conversation with @bungeman Ben points at a certain restriction when using a font as an outline font and COLRv1 font at the same time. I.e. displaying a COLRv1 font in an environment that does not understand COLRv1.

If the bounding box is always retrieved from the base glyph's entry in the `glyf` table, this fallback contour glyph would always need to be at least as wide as its COLRv1 rendering result (potentially widened by invisible points if needed).

Font makers would have more flexibility with the fallback `glyf` glyphs if we had a way to point to a dedicated bounding box glyph separately, other than always referring to the base glyph's `glyf` entry.

rsheeter:
Strawman for a fix for #251. Costs 16 bytes per BaseGlyphPaintRecord ... but for a font that doesn't _have_ an alternate presentation you get some of them back by not writing out bogus glyph records just to express bboxes (as our exploratory compiler does today).

Edit: Fixes #251

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1144,8 +1144,9 @@ For each color glyph, the bounding box for the color glyph is defined within
 the color glyph description (see 5.7.11.2.3). The bounding box data can be
 variable within a variable font.
 
-NOTE: The bounding box of the BaseGlyphPaintRecord can be used to allocate a drawing
-surface without needing to traverse the graph of the color glyph definition.
+NOTE: The bounding box data for the color glyph can be used to allocate a
+drawing surface without needing to traverse the graph of the color glyph
+definition.
 
 A valid color glyph definition shall define a bounded region—that is, it shall
 paint within a region for which a finite bounding box could be defined. The
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1140,7 +1140,9 @@ used for the color glyph. The advance width and height of glyphs referenced by
 PaintGlyph tables are not required to be the same as that of the base glyph and
 are ignored.
 
-The BaseGlyphPaintRecord bbox provides the bounding box for the color glyph.
+For each color glyph, the bounding box for the color glyph is defined within
+the color glyph description (see 5.7.11.2.3). The bounding box data can be
+variable within a variable font.
 
 NOTE: The bounding box of the BaseGlyphPaintRecord can be used to allocate a drawing
 surface without needing to traverse the graph of the color glyph definition.
@@ -1439,16 +1441,6 @@ in COLR version 0, providing records that map a base glyph to a color glyph
 definition. The color glyph definitions that each refer to are significantly
 different, however—see 5.7.11.1.
 
-*VarBBox table:*
-
-| Type | Name | Description |
-|-|-|-|
-| FWORD | xMin | Minimum x of bounding box. For variation, use varIndexBase + 0. |
-| FWORD | yMin | Minimum y of bounding box. For variation, use varIndexBase + 1. |
-| FWORD | xMax | Maximum x of bounding box. For variation, use varIndexBase + 2. |
-| FWORD | yMax | Maximum y of bounding box. For variation, use varIndexBase + 3. |
-| uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
-
 *BaseGlyphList table:*
 
 | Type | Name | Description |
@@ -1470,6 +1462,23 @@ table (5.2.6).
 The records in the baseGlyphPaintRecords array shall be sorted in increasing
 glyphID order. It is intended that a binary search can be used to find a
 matching BaseGlyphPaintRecord for a specific glyphID.
+
+A VarBBox subtable is used to provide the bounding box for the given color
+glyph.
+
+*VarBBox table:*
+
+| Type | Name | Description |
+|-|-|-|
+| FWORD | xMin | Minimum x of bounding box. For variation, use varIndexBase + 0. |
+| FWORD | yMin | Minimum y of bounding box. For variation, use varIndexBase + 1. |
+| FWORD | xMax | Maximum x of bounding box. For variation, use varIndexBase + 2. |
+| FWORD | yMax | Maximum y of bounding box. For variation, use varIndexBase + 3. |
+| uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
+
+The same structure is used in both variable and non-variable fonts. For
+variable data, a base/sequence scheme is used to index into variation mapping
+data. See 5.7.11.4 for details.
 
 The paint table referenced by the BaseGlyphPaintRecord is the root of the graph
 for a color glyph definition.
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1443,10 +1443,10 @@ different, however—see 5.7.11.1.
 
 | Type | Name | Description |
 |-|-|-|
-| UFWORD | x | x-coord of bounding box. For variation, use varIndexBase + 0. |
-| UFWORD | y | x-coord of bounding box. For variation, use varIndexBase + 1. |
-| UFWORD | w | width of bounding box. For variation, use varIndexBase + 2. |
-| UFWORD | h | height of bounding box. For variation, use varIndexBase + 3. |
+| FWORD | xMin | Minimum x of bounding box. For variation, use varIndexBase + 0. |
+| FWORD | yMin | Minimum y of bounding box. For variation, use varIndexBase + 1. |
+| FWORD | xMax | Maximum x of bounding box. For variation, use varIndexBase + 2. |
+| FWORD | yMax | Maximum y of bounding box. For variation, use varIndexBase + 3. |
 | uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
 
 *BaseGlyphList table:*
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1140,11 +1140,9 @@ used for the color glyph. The advance width and height of glyphs referenced by
 PaintGlyph tables are not required to be the same as that of the base glyph and
 are ignored.
 
-The bounding box of the base-glyph contours is used as the bounding box of the
-color glyph. A &#39;glyf&#39; entry with two points at diagonal extrema is
-sufficient to define the bounding box.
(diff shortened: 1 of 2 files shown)


rsheeter:
Another interesting benefit: bbox sharing Just Works, for things with repetitive bboxes (emoji, icons) that's a win.

OFF_AMD2_WD.md
```diff
@@ -1444,6 +1442,16 @@ in COLR version 0, providing records that map a base glyph to a color glyph
...
 different, however—see 5.7.11.1.
 
+*VarBBox table:*
+
+| Type | Name | Description |
+|-|-|-|
+| UFWORD | x | x-coord of bounding box. For variation, use varIndexBase + 0. |
+| UFWORD | y | x-coord of bounding box. For variation, use varIndexBase + 1. |
+| UFWORD | w | width of bounding box. For variation, use varIndexBase + 2. |
+| UFWORD | h | height of bounding box. For variation, use varIndexBase + 3. |
```
justvanrossum, line 1452: Any particular reason to use width/height instead of max_x/max_y? Intuitively I think absolute coordinates are easier to deal with in the context of variations.

PeterConstable: I agree: I'd use xMin, yMin, xMax, yMax to match what is done in the 'glyf' and 'head' tables.

OFF_AMD2_WD.md
```diff
@@ -1444,6 +1442,16 @@ in COLR version 0, providing records that map a base glyph to a color glyph
 definition. The color glyph definitions that each refer to are significantly
 different, however—see 5.7.11.1.
 
+*VarBBox table:*
+
+| Type | Name | Description |
+|-|-|-|
+| UFWORD | x | x-coord of bounding box. For variation, use varIndexBase + 0. |
```
PeterConstable, line 1449: These need to be FWORD, not UFWORD.

OFF_AMD2_WD.md
```diff
@@ -1457,6 +1465,7 @@ different, however—see 5.7.11.1.
 |-|-|-|
 | uint16 | glyphID | Glyph ID of the base glyph. |
 | Offset32 | paintOffset | Offset to a Paint table. |
+| Offset32 | bboxOffset | Offset to a VarBBox table. |
```
PeterConstable, line 1468: I'd state explicitly that bboxOffset can be NULL, in which case the bbox data is taken from the base glyph data. I don't see any downside of allowing that for cases in which that would be adequate, and it can save 12 bytes per colour glyph.

colr-gradients-spec.md
```diff
@@ -525,10 +525,19 @@ struct PaintComposite
   Offset24<Paint>        backdrop;
 };
 
+struct VarBBox {
+  UFWORD                 x; // VarIdx varIndexBase + 0.
```
PeterConstable, line 529: Type should be FWORD. Change members to xMin, yMin, xMax, yMax.

PeterConstable (changes requested):

OFF_AMD2_WD.md
rsheeter: Good catch, fixing

OFF_AMD2_WD.md
rsheeter: Apparently somewhere along the line x,y,w,h stuck in my head as the definition of box :) Agreed it's best to match existing practice.

OFF_AMD2_WD.md
rsheeter: I waffle a bit on that one. Thinking "aloud", two reasons to potentially not do that:

1. Simplicity. Bbox is trivially accessible for both glyf and cff with COLR.
1. Discussion on #251 where a pretty expert group of users expressed concern with use of base glyph

OFF_AMD2_WD.md
rsheeter: Done

OFF_AMD2_WD.md
rsheeter: Done

colr-gradients-spec.md
rsheeter: Done

OFF_AMD2_WD.md
rsheeter: Would it only be true for `glyf` that bbox as described in https://docs.microsoft.com/en-us/typography/opentype/spec/glyf is sufficient or is that true of CFF as well?

PeterConstable: >Simplicity. Bbox is trivially accessible for both glyf and cff with COLR.

Not clear to me what you're saying and why it argues against optionality.

> Discussion on [issue 251...] where a pretty expert group of users expressed concern with use of base glyph

@raphlinus brought up something that I think is a non-issue. I don't There are working implementations of COLR v1 now that have successfully used bbox info from 'glyf'. Those work without issue.

PeterConstable: >Would it only be true for glyf that bbox as described in [OT 'glyf' spec] is sufficient or is that true of CFF as well?

The concept of the bounding box described in the 'glyf' chapter — derived from point coordinate data for on- or off-curve points — would apply equally to CFF(2) as well as to 'glyf'. However, whereas in the 'glyf' table the glyph entries include the bbox values (xMin, etc.), I don't think there's an equivalent in CFF(2) entries. So, a bbox would have to be computed by processing the charstring data.

We could say that a bbox table _should_ (or _must_ / _shall_) be used for fonts that use CFF(2) for base glyphs.

colr-gradients-spec.md
```diff
@@ -525,10 +525,19 @@ struct PaintComposite
   Offset24<Paint>        backdrop;
 };
 
+struct VarBBox {
```
PeterConstable, line 528: Since the same struct will be used for both variable and non-variable cases, is "Var" needed in the name? (We add "Var" in other structs when we're distinguishing from a corresponding non-variable struct.)

rsheeter: I was going for consistency in that Var* would be a variable thing. Don't really have strong feelings about it.

OFF_AMD2_WD.md
rsheeter: I prefer not to have a case where you process the CFF or glyf glyph to get the box. To achieve that I believe I need  it for CFF, and I need it for a variable glyf, and I need it for a glyf with an alternate representation, so we'd be making it optional solely when  there is a glyf table and the bbox desired exists in glyf. And it should be fairly efficient given it's sharable. That leaves me thinking having it mandatory makes sense. WDYT?

colr-gradients-spec.md
rsheeter: Per IM discussion, leaving VarBBox

PeterConstable (approved):

PeterConstable:
@googlebot I consent

OFF_AMD2_WD.md
rsheeter: Resolving per IM and PR approval w/bbox mandatory

anthrotype:
> The same structure is used in both variable and non-variable fonts.

why not having distinct BBox and VarBBox structs and save the unused 4 bytes?

PeterConstable:
How would you determine which structure to read? Have a VarBaseGlyphPaintRecord (and whatever else that might entail)? Have the format for the (Var)BBox table conditioned by whether COLR.itemVariationStoreOffset == NULL?

anthrotype:
ah good point, I had not thought about that. I'm not sure actually.
Checking if VarStore != NULL to determine whether to parse static BBox vs VarBBox would mean that if any color glyph has some COLR variations (and thus has a non-null VarStore), then all the color glyphs (even those that actually do not vary) will use VarBBox.
Perhaps adding a uint8 format to BaseGlyphPaintRecord that says whether it can vary or not (which selects the appropriate BBox or VarBBox), but then we'd spend 1 byte to save 4 so the net effect would be 3 bytes saved overall for static-only fonts, and +1 byte for each color glyphs that can vary in a color VF.
But I'm not sure if it's actually worth it.

rsheeter:
After sanity checking we actually get quite a lot of unique bboxes (text fonts just for comparison, they aren't COLR):

```
twemoji-glyf_colr_1.ttf 10622 unique bboxes
noto-glyf_colr_1.ttf 39370 unique bboxes
Roboto[wdth,wght].ttf 1058 unique bboxes
Arimo[wght].ttf 2503 unique bboxes
```

Maybe worthwhile indeed.

behdad:
I think merging this was premature.

What happens when the paint tries to paint outside the bbox? I don't see that specified. And if not specified, we are bound to see fonts that have bogus bounding box info.

rsheeter:
I don't think this changes that concern? - we already said you should be
able to use a provided bbox, this merely changes where that lives

behdad:
> I don't think this changes that concern? - we already said you should be able to use a provided bbox, this merely changes where that lives

What bothers me most is that this is not even optional. The byte cost cannot be avoided by an implementation that wouldn't need it.

What should have been done instead was to address Ben's concerns by specing out how to calculate bounding box on the fly. This is a very well-known problem. You simply have to run the paint graph / draw twice: first to calculate the bounds so you can allocate, second to do the drawing.

What has happened here is typical of when implementors and spec writers are too close. Things go into the spec without being fully justified. Variations, indeed, as have been pointed out multiple times, make it *impossible* to even specify bounding boxes.

Even before PaintRotate, the *tight* bounding box of an OpenType 1.8 Variable Font was *not* expressible as a variable value using the same variable model.  It *was* possible to specify a non-tight box easily-enough though.

With PaintRotate, even that is not doable. So, to address a concern about rasterization time, one that is well-understood and easily solvable, you have now encoded requirements for the compiler, that we have no idea how to implement.

Indeed, it was fine closing https://github.com/googlefonts/colr-gradients-spec/issues/289 because an implementation did NOT have to use the encoded bbox. But now we are explicitly forcing compilers to compute that, something we don't have a known way to do, short of walking every glyph outline point and calculating it's max bounds under the PaintGraph.

In short, you've replaced a tractable problem with an intractable one.

behdad:
The same thing happened at OTVar1.8 time: MS insisted that bbox variations must be encoded explicitly and forced it through in HVAR/VVAR tables despite objections; only later it was realized that those cannot be encoded given the existing variation model. It's still hanging over our head. Why is that issue being ignored, even in the face of the more problematic PaintRotate?

Issue #319, opened 2021-07-12
Update DeltaSetIndexMap to beyond 16bit


behdad:
Currently `DeltaSetIndexMap` is limited to 16bit entries:

https://docs.microsoft.com/en-us/typography/opentype/spec/hvar

Type | Name | Description
-- | -- | --
uint16 | entryFormat | A packed field that describes the compressed representation of delta-set indices. See details below.
uint16 | mapCount | The number of mapping entries.
uint8 | mapData[variable] | The delta-set index mapping data. See details below.

The `entryFormat` has lots of reserved bits:

Mask | Name | Description
-- | -- | --
0x000F | INNER_INDEX_BIT_COUNT_MASK | Mask for the low 4 bits, which give the count of bits minus one that are used in each entry for the inner-level index.
0x0030 | MAP_ENTRY_SIZE_MASK | Mask for bits that indicate the size in bytes minus one of each entry.
0xFFC0 | Reserved | Reserved for future use — set to 0.

We should use one to add a 32bit variation. To keep it backwards compatible, the 32bit extension should append more data at the end of the current 16bit array. That also enables us to rethink different encoding options, if desirable. For example, a 20bit mode.

I'm thinking about this now.

PeterConstable:
>Currently DeltaSetIndexMap is limited to 16bit entries...
> We should use one to add a 32bit variation...

Please clarify: Do you meant to allow the number of (logical) mapping entries to be >65535?


> To keep it backwards compatible, the 32bit extension should append more data at the end of the current 16bit array.

So, you would keep mapCount as uint16? How would the higher count be represented? 

And is back compat really a problem for this structure? (Existing implementations use it only in HVAR and VVAR, and as long as glyph IDs remain 16-bit there would not be any need in HVAR or VVAR for more than 65535 entries. In COLR, any implementations would be new.) Why not add a flag in entryFormat that determines whether mapCount is uint16 or uint32?

behdad:
> Please clarify: Do you meant to allow the number of (logical) mapping entries to be >65535?

Yes.

> And is back compat really a problem for this structure? (Existing implementations use it only in HVAR and VVAR, and as long as glyph IDs remain 16-bit there would not be any need in HVAR or VVAR for more than 65535 entries. In COLR, any implementations would be new.)

That's a *very* good point. Let's do that.

> Why not add a flag in entryFormat that determines whether mapCount is uint16 or uint32?

[...]

PR #328, opened 2021-07-20
Rev DeltaSetIndexMap for formats 0 and 1

PeterConstable:
Change for #319.

This is a little involved since it requires pulling description of the DeltaSetIndexMap out of the HVAR section into the OT Var Common Formats section (as well as re-defining the structure and adding format 1).

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -9,6 +9,8 @@ Changes to the following sections of ISO/IEC 14496-22:2019 Open Font Format
 - [7.1.7 Algorithm for interpolation of instance values](#changes-to-OFF-717-algorithm-for-interpolation-of-instance-values)
 - [7.2.1 Overview (Font variations common table formats)](#changes-to-off-721-overview-font-variations-common-table-formats)
 - [7.2.3 Item variation stores](#changes-to-off-723-item-variation-stores)
+- [7.3.5 HVAR – Horizontal metrics variations table](#changes-to-off-735---horizontal-metrics-variation-table)
+- [7.3.8 VVAR – Vertical metrics variations table](#changes-to-off-738---vertical-metrics-variation-table)
 - [Bibliography](#changes-to-off-bibliography)
 
 ## Changes to OFF 4.3 Data types
@@ -1298,9 +1300,9 @@ Various table and record formats are defined for COLR version 0 and version 1.
 Several values contained within the version 1 formats are variable. For items
 that vary in a variable font, the variation data is contained in an
 ItemVariationStore table (7.2.3). To associate each variable item with the
-corresponding variation data, a DeltaSetIndexMap table is used, as in the HVAR
-table (7.3.5). Within a given table that has variable items, a base/sequence
-scheme is used to index into the mapping data. See 5.7.11.4 for details.
+corresponding variation data, a DeltaSetIndexMap table (7.2.3.1) is used. Within
+a given table that has variable items, a base/sequence scheme is used to index
+into the mapping data. See 5.7.11.4 for details.
 
 All table offsets are from the start of the parent table in which the offset is
 given, unless otherwise indicated.
@@ -1352,7 +1354,7 @@ The ItemVariationStore is used in conjunction with a BaseGlyphList and its
 subtables, but only in variable fonts. If it is not used, set
 itemVariationStoreOffset to NULL.
 
-The DeltaSetIndexMap format is described in 7.3.5.2. A DeltaSetIndexMap is used
+The DeltaSetIndexMap table is described in 7.2.3.1. Within the COLR table, either format 0 or format 1 of the DeltaSetIndexMap can be used. A DeltaSetIndexMap is used
 in conjunction with the ItemVariationStore in a variable font. The
 DeltaSetIndexMap is optional: if an ItemVariationStore is present but a 
 DeltaSetIndexMap is not included (varIndexMapOffset is NULL), then an implicit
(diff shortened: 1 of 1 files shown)


PeterConstable:
As it might be a bit hard to follow the amendment details ("revise the nth paragraph..."), I've update the [OT 1.9 alpha](https://docs.microsoft.com/en-us/typography/opentype/otspec190alpha/ot190alpha) as well for this change.

OFF_AMD2_WD.md
```diff
@@ -2844,6 +2900,105 @@ flag is not set, or 2 x that amount if the flag is set.
...
+In general, variation deltas are (logically) signed 16-bit integers, and in most
+cases, they are applied to signed 16-bit values (FWORDs) or unsigned 16-bit
+values (UFWORDs). In the COLR table, however, scaled deltas can be applied to
+F2DOT14 or Fixed items, which are fixed-size floating types.
+
+When applying scaled deltas to an F2DOT14 value, the F2DOT14 value is treated
+like a 16-bit integer. (In this sense, the delta and the F2DOT14 value can be
+viewed as integer values in units of 1/16384ths.) If the context in which the
+F2DOT14 is used constrains the valid range for the default value, then any
+variations by applying deltas are clipped to that range.
```
rsheeter, line 2912: I think we phrased "clipped to that range" as "Saturation arithmetic shall be used" elsewhere, perhaps here too?

OFF_AMD2_WD.md
```diff
@@ -2844,6 +2900,105 @@ flag is not set, or 2 x that amount if the flag is set.
...
+for side bearings is recommended. If variation data for side bearings is
+provided, it should include data for both left and right side bearings, and
+mapping subtables for left and right side bearings shall also be included.
+
+_In subclause 7.3.5.2, after the table defnining the horizontal metrics
+variation table, replace the second and third paragraphs with the following:_
+
+Mapping subtables are represented using a DeltaSetIndexMap table; see 7.3.2.1
+for the definition of this table. In the HVAR table, only format 0 of the
+DeltaSetIndexMap is used. Mapping subtables are optional. If a given mapping
```
rsheeter, line 2945: While I cannot immediately see why one would want a format 1 DeltaSetIndexMap in HVAR until we have > 16 bit gid I don't think we need to forbid it. An implementation that supports format1 maps might as well accept them anywhere maps occur.

OFF_AMD2_WD.md
```diff
@@ -2844,6 +2900,105 @@ flag is not set, or 2 x that amount if the flag is set.
...
+## Changes to OFF 7.3.8 - Vertical metrics variation table
+
+_In subclause 7.3.8.2, after the table defining the Vertical metrics variation
+table, replace the remaining text of the subclauase with the following:_
+
+The item variation store table is documented in subclause 7.2.
+
+Mapping subtables are represented using a DeltaSetIndexMap table; see 7.2.3.1
+for the definition of this table. In the VVAR table, only format 0 of the
+DeltaSetIndexMap is used. Mapping subtables are optional. If a given mapping
```
rsheeter, line 2985: Same comment as HVAR

rsheeter (approved):
LGTM, ideally would like @anthrotype to glance over as I know he's spent some time in DeltaSetIndexMap recently for FontTools and may have developed a keener appreciation for it's nuances

rsheeter:
> I've update the OT 1.9 alpha as well for this change.

That's much nicer than trying to count to Nth paragraph in OFF, ty

OFF_AMD2_WD.md
PeterConstable: Back compat: Existing HVAR implementations wouldn't be able to handle fonts that have an HVAR with a format 1 DSIM, and there's no benefit to having an HVAR with a format 1 DSIM at this time. So, it makes best sense to limit HVAR and VVAR to format 0.

OFF_AMD2_WD.md
PeterConstable: Do they mean the same?

This is existing wording; it just moved within the chapter. So I was leaving the wording the same.

rsheeter: If that's just content moving unchanged then +1 to leave it as is.

PeterConstable: Moreover, they don't mean the same, and the intent here is different from the other context in which "saturation arithmetic" was mentioned.

OFF_AMD2_WD.md
anthrotype: yes, until we have 32-bit glyph indices, there's no point in having a format 1 DSIM in {H,V}VAR. The only effect one may achieve in using that is making the font _not_ work on current implementations that don't support format 1.
Perhaps you could change the wording to 

> In the HVAR table, only format 0 of the DeltaSetIndexMap is _currently_ used

implying that in the future (when we get 32-bit gids) this may change

anthrotype (approved):

OFF_AMD2_WD.md
```diff
@@ -2844,6 +2900,105 @@ flag is not set, or 2 x that amount if the flag is set.
...
+Optional mapping subtables can also be used to provide delta-set indices for
+glyph side bearings. In variable fonts with TrueType outlines, variation data
+for side bearings is recommended. If variation data for side bearings is
+provided, it should include data for both left and right side bearings, and
+mapping subtables for left and right side bearings shall also be included.
+
+_In subclause 7.3.5.2, after the table defnining the horizontal metrics
+variation table, replace the second and third paragraphs with the following:_
+
+Mapping subtables are represented using a DeltaSetIndexMap table; see 7.3.2.1
```
vlevantovsky, line 2943: Typo - this reference should be to 7.2.3.1

OFF_AMD2_WD.md
```diff
@@ -1298,9 +1300,9 @@ Various table and record formats are defined for COLR version 0 and version 1.
 Several values contained within the version 1 formats are variable. For items
 that vary in a variable font, the variation data is contained in an
 ItemVariationStore table (7.2.3). To associate each variable item with the
-corresponding variation data, a DeltaSetIndexMap table is used, as in the HVAR
-table (7.3.5). Within a given table that has variable items, a base/sequence
-scheme is used to index into the mapping data. See 5.7.11.4 for details.
+corresponding variation data, a DeltaSetIndexMap table (7.2.3.1) is used. Within
+a given table that has variable items, a base/sequence scheme is used to index
```
vlevantovsky, line 1304: With this change, we added the new definitions of the DeltaSetIndexMap table formats in 7.2.3.1; however, we seem to have done nothing to the already existing parts of the spec (in 7.3.5.2) where the DeltaSetIndexMap table used to be (and still is) defined. Something need to happen here - either we need to make these changes in the same subclause 7.3.5.2, or we need additional amendments to avoid defining the same data table twice.

OFF_AMD2_WD.md
```diff
@@ -2844,6 +2900,105 @@ flag is not set, or 2 x that amount if the flag is set.
...
+
+_In subclause 7.3.5.3, replace the second paragraph with the following:_
+
+Delta-set indices are obtained based on the glyph ID. If there is no delta-set
+index mapping table for advance widths, then glyph IDs implicitly provide the
+indices: for a given glyph ID, the delta-set outer-level index is zero, and the
+glyph ID is the delta-set inner-level index. If delta-set index mappings are
+provided, glyph IDs are used to lookup a mapping entry, which provides the
+outer- and inner-level delta-set indices in a packed format. See 7.3.2.1 for
+details.
```
vlevantovsky, line 2966: Same typo here, the reference should be to 7.2.3.1

OFF_AMD2_WD.md
PeterConstable: Thanks. Fixed in #331

OFF_AMD2_WD.md
PeterConstable: Thanks. Fixed in #331

PR #329, opened 2021-07-21
Provide an optional clip box instead of a mandatory bbox

### LINKED ISSUE #251: Consider allowing indirection to retrieve bounding box from different glyf glyph

rsheeter:
Fixes #251. Again :)

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1538,7 +1538,7 @@ that it not be unnecessarily large.
 | FWORD | yMax | Maximum y of bounding box. For variation, use varIndexBase + 3. |
 | uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
 
-When computing a variable ClipBox compute the min/max coordinates using floating point
+NOTE: When computing a variable ClipBox compute the min/max coordinates using floating point
 values and then round to int values such that the clip box expands. That is, round xMin
 and yMin down and round xMax and yMax up.
 
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1143,8 +1143,11 @@ PaintGlyph tables are not required to be the same as that of the base glyph and
 are ignored.
 
 A valid color glyph definition shall define a bounded region—that is, it shall
-paint within a region for which a finite bounding box could be defined. The
-different paint formats have different boundedness characteristics:
+paint within a region for which a finite bounding box could be defined. A clip
+box can be specified to set overall bounds for a color glyph (see below).
+Otherwise, boundedness is determined by the graph of paint tables that describe
+the color glyph content. The different paint formats have different boundedness
+characteristics:
 
 * PaintGlyph is inherently bounded.
 * PaintSolid, PaintVarSolid, PaintLinearGradient, PaintVarLinearGradient,
@@ -1163,13 +1166,15 @@ details.
 
 A ClipBox table (5.7.11.2.3) may be associated with a color glyph to define
 an overall bounds for the color glyph. The clip box may vary in a variable
-font. If no ClipBox table is present but a bounding box is required by the
-implementation, it shall be computed for a given color glyph by traversing the
-graph of Paint tables that defines that color glyph.
+font.
 
 NOTE: If present, the clip box for a color glyph can be used to allocate a
 drawing surface without needing to traverse the graph of the color glyph
 definition.
+
+NOTE: If no ClipBox table is present but a bounding box is required by the
+implementation, it can be computed for a given color glyph by traversing the
+graph of Paint tables that defines that color glyph.
 
 If a clip box is provided for a color glyph, the color glyph is bounded, and no
 inspection of the Paint graph is required to determine boundedness. If no clip
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -1142,11 +1142,11 @@ used for the color glyph. The advance width and height of glyphs referenced by
 PaintGlyph tables are not required to be the same as that of the base glyph and
 are ignored.
 
-For each color glyph, the bounding box for the color glyph is defined within
-the color glyph description (see 5.7.11.2.3). The bounding box data can be
-variable within a variable font.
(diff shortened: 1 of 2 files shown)


OFF_AMD2_WD.md
```diff
@@ -1140,18 +1140,20 @@ used for the color glyph. The advance width and height of glyphs referenced by
 PaintGlyph tables are not required to be the same as that of the base glyph and
 are ignored.
 
-For each color glyph, the bounding box for the color glyph is defined within
-the color glyph description (see 5.7.11.2.3). The bounding box data can be
-variable within a variable font.
+For each color glyph, a clip box may be associated (see 5.7.11.2.3). The clip
+box may vary in a variable font. If no clip is present it must be computed
+by traversing the graph of Paint that describes the contents of the glyph.
```
drott, line 1145:
```suggestion
box may vary in a variable font. If no clip is present if a bounding box is required by the implementation, it must be computed by traversing the graph of Paint tables that describes the contents of the glyph.
```

OFF_AMD2_WD.md
```diff
@@ -1526,6 +1504,63 @@ are stacked above the top layer of this element.
 Offsets for paint tables not referenced by any PaintColrLayers table should not
 be included in the paintOffsets array.
 
+*ClipBox table:*
+
+The ClipBox table is used to provide a precomputed clip box for the given color glyph.
+Two formats are defined: format 0 for clip boxes without variation, and
+format 1 allowing for clip boxes that can vary in a variable font.
+
+Any content drawn outside the clip box should not render.
```
drott, line 1513:
```suggestion
Any content drawn outside the clip box shall not render.
```
shall, as in MUST  in RFCs?

OFF_AMD2_WD.md
```diff
@@ -1526,6 +1504,63 @@ are stacked above the top layer of this element.
...
+| uint8 | format | Set to 1. |
+| FWORD | xMin | Minimum x of bounding box. For variation, use varIndexBase + 0. |
+| FWORD | yMin | Minimum y of bounding box. For variation, use varIndexBase + 1. |
+| FWORD | xMax | Maximum x of bounding box. For variation, use varIndexBase + 2. |
+| FWORD | yMax | Maximum y of bounding box. For variation, use varIndexBase + 3. |
+| uint32 | varIndexBase | Base index into DeltaSetIndexMap. |
+
+When computing a variable ClipBox compute the min/max coordinates using floating point
+values and then round to int values such that the clip box expands. That is, round xMin
+and yMin down and round xMax and yMax up.
```
drott, line 1542:
```suggestion
Note: When computing a variable ClipBox compute the min/max coordinates using floating point
values and then round to int values such that the clip box expands. That is, round xMin
and yMin down and round xMax and yMax up.
```

drott (changes requested):
Mostly looks good to me. It'd be good to update the pseudocode as well:

> Allocate a bitmap for the glyph according to extents of base glyph contours for gid

could change to something like:

> Allocate a bitmap for the glyph according to the bounding box defined in a matching entry found from ClipList. If no entry is found, perform a glyph traversal to determine maximum bounds.

rsheeter:
> update the pseudocode as well

Good catch, ty. Done.

PeterConstable:
@googlebot I consent.

OFF_AMD2_WD.md
```diff
@@ -1140,19 +1140,10 @@ used for the color glyph. The advance width and height of glyphs referenced by
...
-
-NOTE: If present, the clip box for the color glyph can be used to allocate a
-drawing surface without needing to traverse the graph of the color glyph
-definition.
-
 A valid color glyph definition shall define a bounded region—that is, it shall
 paint within a region for which a finite bounding box could be defined. The
 different paint formats have different boundedness characteristics:
 
-* If a clip box is provided the graph is inherantly bounded, no inspection
```
PeterConstable: The _graph_ isn't bounded. We're asserting that a clipbox makes the color glyph bounded.

PeterConstable: Also, moved out of that list since the list is about paint tables.

OFF_AMD2_WD.md
```diff
@@ -1168,19 +1159,31 @@ sub-graph is bounded.
...
+definition.
+
+If a clip box is provided for a color glyph, the color glyph is bounded, and no
+inspection of the Paint graph is required to determine boundedness. If no clip
+box is defined for a color glyph, however, applications shall confirm that the
+color glyph definition is bounded, and shall not render the color glyph if the
+defining graph is not bounded.
 
 To ensure that rendering implementations do not clip any part of a color glyph,
-the bounding box of the base glyph needs to be large enough to encompass the
```
PeterConstable, line 1174: No longer need to talk about the bbox of the base glyph.

PeterConstable:
Added a commit with various editorial changes. In 5.7.11.2.3, I reversed the order in which the list/record/table structs are presented to match what is done elsewhere. Also changed the clip record range field names to match what is done elsewhere.

PeterConstable:
@googlebot I consent.

PeterConstable:
@googlebot I consent.

PeterConstable:
@googlebot I consent.

OFF_AMD2_WD.md
rsheeter: Good catch.

PeterConstable (approved):

OFF_AMD2_WD.md
rsheeter: I believe you fixed this (ty) so resolving

Issue #339, opened 2021-07-24
Distinguish ItemVariationStore formats (add format 1) due to change to wordDeltaCount field


PeterConstable:
@anthrotype suggested that the change in ItemVariationData of treating the wordDeltaCount field as a packed field with a new meaning assigned to the high-order bit should result in a format bump to ItemVariationStore.

It does seem that should be done:

Currently, the high-order bit of wordDeltaCount should never be set since no existing font would define that many regions (which are roughly proportional to design masters), and in OT 1.8.4 the VariationRegionList.regionCount field was restricted to <32K. That entails that the number of logical entries in each delta set must be <32K. But it probably isn't safe to assume that implementations would validate the ItemVariationData.wordDeltaCount field is, accordingly, <32K.

If existing implementations read a format 0 ItemVariationStore in which the high-order bit of wordDeltaCount is set, they may assume there are >32K in16 entries in each delta set. That would likely result in wrong interpretation of the data, but could also result in read overrun past the end of the ItemVariationStore (and potential security vulnerability).

Separate formats 0 and 1 can be defined: they would have the same fields, but in format 0 the LONG_WORDS flag is not used.

anthrotype:
the current format is 1 so the proposed new format would be format 2

https://docs.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#item-variation-store-header-and-item-variation-data-subtables

justvanrossum:
> Separate formats 0 and 1 can be defined: they would have the same fields, but in format 0 the LONG_WORDS flag is not used.

This feels a bit redundant: couldn't the new format be for "long words" only, and the old format for "short words" only?

anthrotype:
the `wordDeltaCount` field (whose high bit becomes the LONG_WORDS flag) is defined on the ItemVariationData subtable, not on the parent ItemVariationStore table, where the Format field is specified. A Format 2 VarStore can contain a mix of VarData subtables, some of which use shorts, other use longs, depending on the LONG_WORDS flag. A compiler would use format 2 whenever there's any VarData subtable that contains long deltas, or the current format 1 when none do.

justvanrossum:
Ah, somehow I missed that completely. Sorry for the noise.

anthrotype:
a Format 2 VarStore containing no "long" VarData is arguably useless but still valid

behdad:
No.

PeterConstable:
> No.

@behdad Please clarify: What are your responding to?

[...]

PR #342, opened 2021-07-26
Add recommendation regarding use of LONG_WORDS flag

PeterConstable:
Recommendation added in relation to #339, plus one editorial addition.

### DIFF
--- a/OFF_AMD2_WD.md
+++ b/OFF_AMD2_WD.md
@@ -2884,7 +2884,15 @@ target items. Conceptually, the deltas form a two-dimensional array, with
 delta-set rows that include a delta for each of the regions referenced by that
 subtable. From this perspective, the delta-set columns correspond to regions.
 
-_Replace the fifth paragraph that follows the figure in (now) 7.2.3.2 ("A
+_Before the fourth paragraph following the figure in (now) 7.2.3.2 ("Note that there is minimal overhead..."), insert the following paragraph:_
+
+Similarly, ItemVariationData subtables can also combine delta values with 32-bit
+and 16-bit representations. An optimizing compiler can organize deltas into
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

OFF_AMD2_WD.md
```diff
@@ -2948,6 +2956,9 @@ int32 and int16 values. If the flag is not set, deltas are presented using a
 mix of int16 and int8 values. See the description of the DeltaSet record
 below for additional details.
 
+The LONG_WORDS flag should only be used in top-level tables that include 32-bit
+values that can be variable—currently, only the COLR table.
```
behdad, line 2960: I don't like this. Puts unnecessary restriction in the spec that needs to be kept uptodate.

I suggest instead you just document that OT <= 1.X didn't support this flag. Let compilers figure out what to do or err depending on what version they target.

rsheeter: +1.

PeterConstable: Saying simply that OT <= 1.9 didn't support this flag would suggest that, from 1.9 on, the flag is valid _everywhere_. But that could well lead to fonts that use the flag in tables like HVAR where, for many existing implementations in the wild, it would not work and could even lead to buffer overruns. I don't see why it's a problem to recommend for font compilers not to do this. 

If there's a concern about having less in the spec to maintain, I don't see a problem dropping "currently, only the COLR table".

behdad: > Saying simply that OT <= 1.9 didn't support this flag would suggest that, from 1.9 on, the flag is valid _everywhere_.

It truly saddens me that I have to reply to nonsensical arguments like this. And you studied math... sighz.

That's not how "backwards compatibility" works.

PR #347, opened 2021-08-02
define ClipList.format (0) to make it more futureproof

anthrotype:
Fixes https://github.com/googlefonts/colr-gradients-spec/issues/346

PR #348, opened 2021-08-03
Rev C++ header for greater consistency

PeterConstable:
Revised some field names for greater consistency internally and with the OT/OFF spec names.

Also revised description of ClipList to use ArrayOf as elsewhere in the C++ header.

### DIFF
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -560,8 +560,7 @@ typedef ArrayOf<Offset32<Paint>, uint32> LayerList;
 struct ClipList
 {
   uint8                 format;  // Set to 1.
-  uint32                numClips;
-  Clip                  clips[/*numClips*/];  // Clip records, sorted by startGlyphID
+  ArrayOf<Clip, uint32> clips;  // Clip records, sorted by startGlyphID
 }
 
 struct COLRv1
--- a/colr-gradients-spec.md
+++ b/colr-gradients-spec.md
@@ -560,7 +560,8 @@ typedef ArrayOf<Offset32<Paint>, uint32> LayerList;
 struct ClipList
 {
   uint8                 format;  // Set to 1.
-  ArrayOf<Clip>         clips;  // Clip records, sorted by startGlyphID
+  uint32                numClips;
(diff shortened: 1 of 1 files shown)


colr-gradients-spec.md
```diff
@@ -561,8 +560,7 @@ typedef ArrayOf<Offset32<Paint>, uint32> LayerList;
 struct ClipList
 {
   uint8                 format;  // Set to 1.
-  uint32                numClips;
-  Clip                  clips[/*numClips*/];  // Clip records, sorted by startGlyphID
+  ArrayOf<Clip>         clips;  // Clip records, sorted by startGlyphID
```
behdad, line 563: `ArrayOf` uses a 16bit count; we had 32 here. Call it `Array32Of`.

@rsheeter do we need more than 16bit though?

anthrotype: previously we had `ArrayOf<Clip, uint32> ClipList`
https://github.com/googlefonts/colr-gradients-spec/pull/347/files#diff-458fde56ab00bcf68739126026367f5098e0ea122f477fbd2ec03d870eba40e7L561

> do we need more than 16bit though?

probably not, but why restrict oneself, if one really wishes to assign one individual clip box to every single base colr glyph and there are > 65k of them.. It's only the count type (16bit vs 32bit), so +2 bytes in total per ClipList table (and there's only one per font)

PeterConstable: Sorry, I had overlooked that ArrayOf uses uint16.

For a count that occurs once, I don't see any reason _not_ to use uint32. I'll restore that part.

behdad:
Agree that keeping 32bit is good.

Feel free to also makes sizes explicit everywhere in this code. In HarfBuzz we now have `Offset16To/Offset24To/Offset32To/Array16Of/Array32Of`.

colr-gradients-spec.md
PeterConstable: I looked again and realized ArrayOf could be used with uint32 as a second arg, so changing to that.

PeterConstable:
If others way to change ArrayOf with optional count type to Array16Of / Array32Of, etc. I have no opinion and will let others do that.

rsheeter (approved):

PR #352, opened 2021-09-06
Fix pseudocode for compositing step

drott:
The pseudocode for compositing was missing one layer allocation. This
would result in incorrectly drawing BackdropPaint just on the previous
state of the canvas, then compositing the source paint down on
everything, instead of compositing backdrop and source first, then
applying the result to the current state of the canvas.


bungeman:
Just wanted to point out that this is equivalent to OFF_AMD2_WD.md 5.7.11.3, which already states that both the source and backdrop drawings are saved, then the two are composed, then the result of the composition blended with the destination.

PR #362, opened 2022-05-12
DO NOT MERGE: Draft for PaintSweepGradient update

drott:


OFF_AMD2_WD.md
```diff
@@ -591,57 +591,57 @@ A sweep gradient is defined by a center point, starting and ending angles, and a
...
+can be sampled for colors outside the defined ColorLine interval.
+
+To draw the sweep gradient, for each position along the circular arc, starting
+from from 0 degrees, ending at 360°, a ray from the center outward is painted
+with the color of color line where the ray intersects with the circular arc at
+that particular angle. Angle positions on the spiraling circular arc below 0°
+and above 360° are not sampled for drawing the rays.
+
+Figure 5.30 illustrates drawing direction from 0° to 360°, start and end angles
+and color stop alignment. The color for angles lower than 0° is red and the
```
niklasb-ms, line 612: I think this is meant to say "angles lower than 30", i.e., the start angle.

OFF_AMD2_WD.md
```diff
@@ -591,57 +591,57 @@ A sweep gradient is defined by a center point, starting and ending angles, and a
...
+using a `ColorLine` with color stops 0 and 1 for red and yellow and extend mode
+pad.**
+
+Not more than one full rotation is drawn and there is no overlap in drawing for
+angles outside the 0° and 360° range. However, start and end angles can be
+positioned at angles below 0° and above 360°. Through that, and through how wide
+the ColorLine interval is defined, color stops may lie outside the 0° to 360°
+circle. This has an effect on the computation of the gradient colors inside the
+interval of 0° to 360°, but colors are not sampled from outside this interval.
+See figure 5.31 for an example. The color for angles lower than 330° is rad due
```
niklasb-ms, line 628: Typo. "rad" should be "red".

OFF_AMD2_WD.md
```diff
@@ -591,57 +591,57 @@ A sweep gradient is defined by a center point, starting and ending angles, and a
...
-red at stop offset 1.0.
-
-![A sweep gradient, from red to yellow to red, with a smooth transition at the start/end angle 0°.](images/colr_conic_gradient_rotation-0.png)
-
-**Figure 5.32 A sweep gradient with a smooth transition at the start/end angle 0°**
+For sampling colors, the color line is aligned to an infinitely spiraling
+circular arc around the center point, with arbitrary radius. The position 0° on
+the arc means the direction of the positive x-axis, 360° means one full
+counter-clockwise rotation.  The ColorLine's stop offset 0 is aligned with the
+starting angle, and stop offset 1 aligned with the ending angle.
```
niklasb-ms, line 598: The spec should say what happens if the ending angle is _less than_ the starting angle.
One possibility is that this is not allowed and nothing is drawn.
The other possibility is that the color line arc spirals in a _clockwise_ direction from the starting angle to the ending angle.

niklasb-ms (changes requested):

OFF_AMD2_WD.md
drott: Done.

OFF_AMD2_WD.md
drott: Indeed, fixed, thanks.

OFF_AMD2_WD.md
niklasb-ms: Yes, thank you.
There is also the question of what to render if startAngle == endAngle.
In this case it's not obvious how to map angles greater or less than the start/end angle to color line positions. For now, I chose to map greater angles to FLT_MAX and lesser angles to FLT_MIN. This produces arguably appropriate results for pad extend mode, but not for other extend modes. What do you propose?

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
-red at stop offset 1.0.
-
-![A sweep gradient, from red to yellow to red, with a smooth transition at the start/end angle 0°.](images/colr_conic_gradient_rotation-0.png)
-
-**Figure 5.32 A sweep gradient with a smooth transition at the start/end angle 0°**
+For sampling and interpolating colors, the color line is aligned to an
+infinitely spiraling circular arc around the center point, with arbitrary
+radius. The position 0° on the arc means the direction of the positive x-axis,
+360° means one full counter-clockwise rotation.  The ColorLine's stop offset 0
+is aligned with the start angle, and stop offset 1 aligned with the end angle,
```
PeterConstable, line 598: "and stop offset 1 **is** aligned"

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+independent of which angle is larger.
+
+NOTE: If the start angle is larger than the end angle, this means that the
```
PeterConstable, line 601: Current spec assumes mod 360, and hence it can assume the color line always progresses in one direction, even when start == stop, With the new semantics, what's the direction if the start and end angles are the same?

I'd word the note as follows:

NOTE: If the start angle is less than or equal to the end angle, the color line progresses from smaller to larger offsets in the counter-clockwise direction. If the start angle is larger than the end angle, however, the color line is inverted and progresses in the clockwise direction.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+Outside the defined interval of the ColorLine, the color value of a position on
+the ColorLine is filled in depending on its extend mode. See 5.7.11.1.2.1 Color
+Lines for more details. In effect, this means that the spiraling circular arc
+can be sampled for colors outside the defined ColorLine interval.
+
+To draw the sweep gradient, for each position along the circular arc, starting
+from from 0 degrees, ending at 360°, a ray from the center outward is painted
+with the color of color line where the ray intersects with the circular arc at
+that particular angle. Angle positions on the spiraling circular arc below 0°
+and above 360° are not sampled for drawing the rays.
```
PeterConstable, line 613: For the edge case color @ 0 != color @ 360, should we say "and above or equal to 360º"?

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+
+Figure 5.30 illustrates drawing direction from 0° to 360°, start and end angles
```
PeterConstable, line 615: Suggested:
"Figure 5.30 illustrates a sweep gradient with colors progressing from 0º to 360º. The gradient is specified with a start angle of 30º and end angle of 150º. The color line is specified with a yellow stop at offset 0 (aligned to the start angle) and a red stop at offset 1 (aligned to the end angle). The pad extend mode is used, hence the color for angles below 30º is yellow, and the color for angles above 150º is red."

images/colr_conic_gradient_start_stop_angles.html
```diff
@@ -0,0 +1,66 @@
+
```
PeterConstable, line 1: The overlays are a bit crowded. Can we make the figure and spread out the overlays? And could we somehow differentiate the indications for start/end angle vs. stop offsets 0 and 1?

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+and color stop alignment. The color for angles lower than 30° is red and the
+color for angles greater than 150° is yellow due to the pad extend mode.
+
+![A sweep gradient, from yellow to red, with start angle of 30° and an end angle
+of 150°.](images/colr_conic_gradient_start_stop_angles.png)
+
+**Figure 5.30 A sweep gradient with start angle of 30° and an end angle of 150°,
+using a `ColorLine` with color stops 0 and 1 for yellow and red and extend mode
```
PeterConstable, line 623: "... and pad extend mode."

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+pad.**
+
+Not more than one full rotation is drawn and there is no overlap in drawing for
+angles outside the 0° and 360° range. However, start and end angles can be
+positioned at angles below 0° and above 360°. Through that, and through how wide
+the ColorLine interval is defined, color stops may lie outside the 0° to 360°
+circle. This has an effect on the computation of the gradient colors inside the
+interval of 0° to 360°, but colors are not sampled from outside this interval.
+See figure 5.31 for an example. The color for angles lower than 330° is red due
```
PeterConstable, line 632: "lower than 330º is red"?? Lower suggests 0 to 330, but yellow is shown in that range. Did you mean to say "yellow"?

OFF_AMD2_WD.md
```diff
@@ -1852,8 +1856,8 @@ formats, see 5.7.11.2.4. For background information on the color line, see
 | Offset24 | colorLineOffset | Offset to ColorLine table. |
 | FWORD | centerX | Center x coordinate. |
 | FWORD | centerY | Center y coordinate. |
-| F2DOT14 | startAngle | Start of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. |
-| F2DOT14 | endAngle | End of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. |
+| F2DOT14 | startAngle | Start of the angular range of the gradient, add 1.0 and multiply by 180° to retrieve counter-clockwise degrees. |
```
PeterConstable, line 1859: Change punctuation to colon (or dash):

"Start of the angular range of the gradient: add..."

Similarly in subsequent occurrences.

OFF_AMD2_WD.md
```diff
@@ -1863,8 +1867,8 @@ formats, see 5.7.11.2.4. For background information on the color line, see
 | Offset24 | colorLineOffset | Offset to VarColorLine table. |
 | FWORD | centerX | Center x coordinate. For variation, use varIndexBase + 0. |
 | FWORD | centerY | Center y coordinate. For variation, use varIndexBase + 1. |
-| F2DOT14 | startAngle | Start of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. For variation, use varIndexBase + 2. |
-| F2DOT14 | endAngle | End of the angular range of the gradient, 180° in counter-clockwise degrees per 1.0 of value. For variation, use varIndexBase + 3. |
+| F2DOT14 | startAngle | Start of the angular range of the gradient, add 1.0 and multiply by 180° to retrieve counter-clockwise degrees. For variation, use varIndexBase + 2. |
```
PeterConstable, line 1870: Adding the bias makes this different from how angles are represented in transforms. Let's confirm that's not a concern. (I'm OK with it as it has good motivation.)

Assuming we agree, it might be worth calling out this difference. Suggest adding a note after line 1877:

> Angles are expressed in counter-clockwise degrees from the direction of the positive x-axis in the design grid.
> NOTE: A bias of 1.0 is used in the representation of start and end angles of sweep gradients. This is not done, however, for angles in rotate and skew transforms (5.7.11.2.5.11 and 5.7.11.2.5.12).

The same sentence can be added in the notes at lines 2253 and 2358.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+
+![A sweep gradient, from yellow to red, with start angle of 330° and an end angle
+of 390°.](images/colr_conic_gradient_stop_angle_outside.png)
+
+**Figure 5.31 A sweep gradient with start angle of 330° and an end angle of 390°,
+using a `ColorLine` with color stops 0 and 1 for yellow and redd and extend mode
+pad.**
+
+Because the sweep gradient is drawn from 0° to 360° a sharp transition may occur
+at 0°, this can be mitigated by adjusting the color stops at the 0° and 360°
```
PeterConstable, line 645: The current spec has a figure illustrating the smooth transition. This draft drops that figure. A consequence is that all of the subsequent 20 figures will need to be re-numbered, which is not done in this draft. Perhaps we keep the extra figure to avoid needing to renumber all the other figures?

PeterConstable (commented):
I've left various comments. I've prepared an alternate HTML for figure 5.30: 

image

GitHub won't let me attach the HTML file to this comment, so I'll find another way.

PeterConstable:
See #363 for suggested alternate image.

OFF_AMD2_WD.md
drott: Added an extra figure to explain the inversion that @niklasb-ms brought up. Smooth transition text also prefixed with `NOTE:` as there are no specific mandatory rules in this paragraph.

OFF_AMD2_WD.md
drott: Done.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+position on the arc to have the same color. The location of the transition axis
+can also be shifted by nesting the PaintSweepGradient inside a
+PaintRotate/PaintVarRotate operation.
```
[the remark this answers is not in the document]
drott: Used the suggested wording.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,65 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+pad.**
+
+Not more than one full rotation is drawn and there is no overlap in drawing for
+angles outside the 0° and 360° range. However, start and end angles can be
+positioned at angles below 0° and above 360°. Through that, and through how wide
+the ColorLine interval is defined, color stops may lie outside the 0° to 360°
+circle. This has an effect on the computation of the gradient colors inside the
+interval of 0° to 360°, but colors are not sampled from outside this interval.
+See figure 5.31 for an example. The color for angles lower than 330° is red due
+to the extend mode pad, then the color line transitions from 330° and red to
```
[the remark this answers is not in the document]
drott: Done.

OFF_AMD2_WD.md
drott: Corrected, thanks for catching that. Missed that in the text after changing the figure.

OFF_AMD2_WD.md
drott: Done.

images/colr_conic_gradient_start_stop_angles.html
drott: I updated the image in the following way - made white background circles around the stop positions to highlight those and separate them from the background. Used different angles to have more space for labels. I experimented with fading the sweep gradient towards white as well to indicate that it should not be considered clipped, but that didn't look so good. I still prefer to avoid clipping it - could you let me know what you think of the new version - are the stop positions clear?

OFF_AMD2_WD.md
drott: Used that text with the modification of 

> with colors progressing from 0º to 360º

"with the drawing direction progressing..."

as I want to highlight that logically that's how the colors are supposed to be sampled and drawn. WDYT?

PeterConstable: SGTM. Btw, I mistakenly entered º (U+00BA) in some cases rather than ° (U+00B0), so watch for that if you copy my text. (00BA will have an underbar in most fonts.)

images/colr_conic_gradient_start_stop_angles.html
PeterConstable: You'll need to push a commit before I can review.

OFF_AMD2_WD.md
drott: We mention below how transforms affect sweeps, and how that can flip direction, hence I would prefer to avoid the notion of clockwise and counter-clockwise. 

Figure 5.31 added, to further illustrate and explain the inversion.

As for start angle == end angle exactly, I am not sure what we should mandate. Extend mode pad works, for the others, I am not sure what we should say or if should spec it.

OFF_AMD2_WD.md
drott: Done.

OFF_AMD2_WD.md
drott: Not sure, maybe we need to discuss this - then we would need to say ending "just before 360°" in the text before.

PeterConstable:
> Are you suggesting to wait with merging until the MPEG AHG has accepted the proposed change, too?

No, that's not what I meant. These revisions are in OFF_AMD2_WD, which was proposed content for Amendment 2 of OFF. But it's too late to make further changes to amendment 2. So, we'll need to prepare a separate document for that.

OFF_AMD2_WD.md
PeterConstable: I agree. We say that angles are counter-clockwise from the +ve x axis in multiple places. Clearly a transform (e.g., vertical mirror) can change the visual result as though the angles were in clockwise direction, but that doesn't keep us from saying that angles are interpreted as counter-clockwise.

The angles determine alignment of stop offsets 0 and 1, and the color line progresses through those offsets as with any color line, That should be enough to be unambiguous. But just in case, we want to make sure no one misunderstands and has the colorline go in the wrong direction. I think the note I suggested above makes that clear.

OFF_AMD2_WD.md
PeterConstable: Thinking out loud:

Let's consider what happens if start/end angles vary and are converging. And let's compare to similar cases for linear and radial.

The comparable issue for linear gradients is that in which p0p1 converges with p0p2. For that case, we say,
>If line p₀p₂ is parallel to line p₀p₁ (or near-parallel for an implementation-determined definition), then the gradient is ill-formed and must not be rendered.
That's a particularly problematic case: line p0p2 represents the direction of iso-colour drawing, but it becomes completely ambiguous which colour is to be drawn on that line, and there is no colour specified anywhere off that line. And this is true regardless of the extend mode. For that case we say _Don't render anything!_

For radial gradients, the comparable issue occurs when the circles converge. For the limiting case in which they are identical, we say, "Nothing is painted." (That case has exactly the same problems as for p0p1 == p0p2.) For cases in which they are _nearly_ identical, things differ between pad mode versus either mirror or reflect modes. For pad mode, you can get a solid colour circle on a solid colour background (the gradient between them being less than one pixel so not seen). We haven't called out that case. But for mirror and reflect modes, we call out the high spatial frequency transitions that arise, and that they will lead to implementation-dependent artefacts like Moiré patterns.


For sweep gradients, I think pad mode is also distinct from mirror and reflect modes. With pad mode, things remain well defined with most/all of the 360 arc will have a solid color. When start == end, I'd expect one colour from 0° to start/end, and the other from start/end to 360°

But with mirror or reflect modes, as start/end converge, the interval of iteration (wavelength) gets smaller, approaching 0, just as for the comparable linear/radial cases. With a really small interval, spatial frequency may be smaller or greater than the pixel size, depending on the distance from the center, I suppose you'd see implementation-dependent artifacts close in, with clear mirror/reflect gradients if you get far enough out from the center. But for the limiting case of start == end, then the interval of iteration becomes 0.

So, I'm inclined to say
- For pad mode, when start == end, a sweep gradient is drawn as two solid colour regions.
- For mirror and extend modes:
  - When start and end are nearly identical there will be implementation-dependent artefacts (similar to what is shown in 5.28 for radial gradients).
  - When start == end, the gradient is ill-formed and nothing is drawn.

images/colr_conic_gradient_start_stop_angles.html
drott: Closing this, I think the additional image files appeared in changed files. Let me know if these can't be seen.

OFF_AMD2_WD.md
drott: Edited, new wording: 

> To draw the sweep gradient, for each position along the circular arc, starting
from from 0° up to but not including 360°, a ray from the center outward is
painted with the color of color line where the ray intersects with the circular
arc at that particular angle. Angle positions on the spiraling circular arc
below 0° and equal to or above 360° are not sampled for drawing the rays.

OFF_AMD2_WD.md
drott: SGTM, thanks for these analogies, added wording to prevent drawing when start == end, and a note regarding artifacts at small angular distance.

OFF_AMD2_WD.md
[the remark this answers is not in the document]
drott: This one is fixed now.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,88 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+is larger than the end angle.**
+
+![A sweep gradient, from yellow to red, with start angle of 330° and an end angle
+of 400°.](images/colr_conic_gradient_stop_angle_outside.png)
+
+**Figure 5.32 A sweep gradient with start angle of 330° and an end angle of 400°,
+using a `ColorLine` with color stops 0 and 1 for yellow and red and extend mode
+pad.**
+
+NOTE: Because the sweep gradient is drawn from 0° to 360° a sharp transition may
```
PeterConstable, line 667: "... a sharp transition may occur at 0°. This can be mitigated ..."

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,88 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
-
-**Figure 5.32 A sweep gradient with a smooth transition at the start/end angle 0°**
+For sampling and interpolating colors, the color line is aligned to an
+infinitely spiraling circular arc around the center point, with arbitrary
+radius. The position 0° on the arc means the direction of the positive x-axis,
+360° means one full counter-clockwise rotation.  The ColorLine's stop offset 0
+is aligned with the start angle, and stop offset 1 is aligned with the end
+angle, independent of which angle is larger.
+
+NOTE: If the start angle is less than or equal to the end angle, the color line
```
PeterConstable, line 601: This note should move after fig 5.31.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,88 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+yellow, and the color for angles above 150° is red.
+
+![A sweep gradient, from yellow to red, with start angle of 110° and an end angle
+of 230°.](images/colr_conic_gradient_start_stop_angles.png)
+
+**Figure 5.30 A sweep gradient with start angle of 110° and an end angle of
+230°, using a `ColorLine` with color stops 0 and 1 for yellow and red and pad
+extend mode.**
+
+Not more than one full rotation is drawn and there is no overlap in drawing for
```
PeterConstable, line 641: Move the note earlier about stop angle > end angle here. (Perhaps drop the "NOTE:" prefix?) Then move Fig 5.32 to directly follow (and re-number to 5.31).

The this paragraph ("Not more than..." follows that figure, with fig 5.31 (re-numbered to 5.32) after that.

PeterConstable (changes requested):
I think some items might need to be reordered. Fig 5.31 and Fig 5.32 are back to back, but 5.32 pertains to the text before 5.31. Fig 5.31 should follow Fig 5.30, but the text that pertains to 5.31 comes earlier. 

I've added comments at relevant points below with suggestions.

OFF_AMD2_WD.md
drott: In the case of figure 5.30, and if I am not mistaken, in other cases as well, we describe textually what happens, then we illustrate with a figure after that, don't we? In other words, shouldn't we have the note before the figure 5.31 (which is the one for inversion) here as well?

I moved the note, so if I am not mistaken, no we have the following order:

* Figure 5.30 illustrates...
* Figure 5.30,
* NOTE: If the start angle is less than or equal to the end angle...see figure 5.31.
* Figure 5.31
* "Not more than one full...see figure 5.32 for an example"
* Figure 5.32.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,88 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+angles outside the 0° and 360° range. However, start and end angles can be
+positioned at angles below 0° and above 360°. Through that, and through how wide
+the ColorLine interval is defined, color stops may lie outside the 0° to 360°
+circle. This has an effect on the computation of the gradient colors inside the
+interval of 0° to 360°, but colors are not sampled from outside this interval.
+See figure 5.32 for an example. The color for angles lower than 330° is yellow
+due to the extend mode pad, then the color line transitions from 330° and yellow
```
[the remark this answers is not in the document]
drott: thanks, done.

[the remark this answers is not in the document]
drott: Done.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,88 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+To draw the sweep gradient, for each position along the circular arc, starting
+from from 0° up to but not including 360°, a ray from the center outward is
+painted with the color of color line where the ray intersects with the circular
+arc at that particular angle. Angle positions on the spiraling circular arc
+below 0° and equal to or above 360° are not sampled for drawing the rays.
+
+If the ColorLine's extend mode is reflect or repeat and start and end angle are
+equal, nothing must be drawn.
+
+NOTE: When the sweep gradient's ColorLine uses the repeat or reflect extend mode
```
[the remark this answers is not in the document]
drott: Edits incorporated.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
-transitions from red at stop offset 0, to yellow at stop offset 0.5, and back to
-red at stop offset 1.0.
-
-![A sweep gradient, from red to yellow to red, with a smooth transition at the start/end angle 0°.](images/colr_conic_gradient_rotation-0.png)
-
-**Figure 5.32 A sweep gradient with a smooth transition at the start/end angle 0°**
+For sampling and interpolating colors, the color line is aligned to an
+infinitely spiraling circular arc around the center point, with arbitrary
+radius. The position 0° on the arc means the direction of the positive x-axis,
+360° means one full counter-clockwise rotation.  The ColorLine's stop offset 0
```
PeterConstable, line 597: "... and 360° means one full counter-clockwise rotation. The color line’s stop offset 0..."

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+is aligned with the start angle, and stop offset 1 is aligned with the end
```
PeterConstable, line 598: "... is aligned with the start angle and stop offset 1 is aligned..."

(delete the comma)

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+angle, independent of which angle is larger.
+
+Outside the defined interval of the ColorLine, the color value of a position on
```
PeterConstable, line 601: This paragraph is repeating information provided in the color line section. We don't repeat this explanation when linear or radial gradients are described; we simply say that at some point the color of the color line is used. I think this paragraph can be deleted.

(In the existing spec, some similar details are explained, but that was motivated by the different semantics for sweep gradients that was being used.)

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+the ColorLine is filled in depending on its extend mode. See 5.7.11.1.2.1 Color
+Lines for more details. In effect, this means that the spiraling circular arc
+can be sampled for colors outside the defined ColorLine interval.
+
+To draw the sweep gradient, for each position along the circular arc, starting
```
PeterConstable, line 606: "To draw the sweep gradient, for each position along the circular arc starting from 0° and up to but not including 360°, a ray from the center outward is painted with the color of the color line where the ray intersects the circular arc at that particular angle. Angle positions below 0° and equal to or above 360° are not sampled for drawing the rays."

OFF_AMD2_WD.md
```diff
@@ -577,7 +577,7 @@ A sweep gradient provides a gradation of colors that sweep around a center
 point. For a given color on a color line, that color projects as a ray from the
```
PeterConstable, line 577: Lines 306 and 307 (note about repeat/mirror modes being special for sweep) need to be deleted.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+angle of 230°. The color line is specified with a yellow stop at offset 0
+(aligned to the start angle) and a red stop at offset 1 (aligned to the end
+angle). The pad extend mode is used, hence the color for angles below 110° is
+yellow, and the color for angles above 150° is red.
+
+![A sweep gradient, from yellow to red, with start angle of 110° and an end angle
+of 230°.](images/colr_conic_gradient_start_stop_angles.png)
+
+**Figure 5.30 A sweep gradient with start angle of 110° and an end angle of
+230°, using a `ColorLine` with color stops 0 and 1 for yellow and red and pad
```
PeterConstable, line 632: "... using a color line..."

PeterConstable (changes requested):
Working on incorporating revisions into OT 1.9.1. Along the way I'm spotting some editorial details.

Partway done; will finish tomorrow.

OFF_AMD2_WD.md
```diff
@@ -1873,6 +1899,10 @@ mapping data; see 5.7.11.4 for details.
 Angles are expressed in counter-clockwise degrees from the direction of the
 positive x-axis in the design grid.
 
+NOTE: A bias of 1.0 is used in the representation of start and end angles of
+sweep gradients. This is not done, however, for angles in rotate and skew
+transforms (5.7.11.2.5.11 and 5.7.11.2.5.12).
```
anthrotype, line 1904: it would be nice to explain why this +1.0 bias is only applied for sweep gradient angles and not for other rotate or skew angles

PeterConstable: Suggestion:
> NOTE: To allow for representation of +360°, a bias of 1.0 is used in the representation of start and end angles of sweep gradients. ...

anthrotype: yet another example, 225 degree angle represented using the biased F2Dot14 angle:

```python
>>> v = 4096
>>> bits = 14
>>> angle = (v / (1 << bits) + 1.0) * 180
>>> print(angle)
225.0
>>> v2 = (angle / 180 - 1.0) * (1 << 14)
>>> assert v2 == v
```

PeterConstable: Here's my draft revision (as porting to OT 1.9.1):

> NOTE: To allow for a representation of +360°, a bias of 1.0 is used in the representation of stand and end angles of sweep gradients. For example, an F2DOT14 value of -2.0 (0x8000) represents -180°; an F2DOT14 value of 0.0 (0x0000) represents +180°; an F2DOT14 value of 0.25 (0x1000) represents +225°; an F2DOT14 value of 1.0 (0x4000) represents +360°. However, a bias is not used for representation of angles in rotate or skew transforms (5.7.11.2.5.11, 5.7.11.2.5.12).

PeterConstable (changes requested):

OFF_AMD2_WD.md
anthrotype: Another option could be that we change the type of the angles from Fixed2Dot14 to a new Fixed4Dot12 for example, always a 16-bit type but it would have slightly smaller fractional part, which would still allow up to 3 decimal digits when converted to/from float, and it could allow to encode values between -16.0 <= v < 15.9998, which in degrees means -2880.0° <= angle < +2879.95605°

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+of 400°.](images/colr_conic_gradient_stop_angle_outside.png)
+
+**Figure 5.32 A sweep gradient with start angle of 330° and an end angle of 400°,
+using a `ColorLine` with color stops 0 and 1 for yellow and red and extend mode
+pad.**
+
+NOTE: Because the sweep gradient is drawn from 0° to 360° a sharp transition may
+occur at 0°. This can be mitigated by adjusting the color stops at the 0° and
+360° position on the arc to have the same color. The location of the transition
+axis can also be shifted by nesting the PaintSweepGradient inside a
```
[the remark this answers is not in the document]
drott: Done.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+interval of 0° to 360°, but colors are not sampled from outside this interval.
+See figure 5.32 for an example. The color for angles lower than 330° is yellow
+due to the pad extend mode, then the color line transitions from yellow at 330°
+to red at 400°, but only the color values up until 360° are sampled for drawing.
+
+![A sweep gradient, from yellow to red, with start angle of 330° and an end angle
+of 400°.](images/colr_conic_gradient_stop_angle_outside.png)
+
+**Figure 5.32 A sweep gradient with start angle of 330° and an end angle of 400°,
+using a `ColorLine` with color stops 0 and 1 for yellow and red and extend mode
```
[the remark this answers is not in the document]
drott: Done.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+line.](images/colr_conic_gradient_inverted.png)
+
+**Figure 5.31 A sweep gradient, from yellow to red, with start angle of 210° and
+end angle of 110°, showing the inversion of the color line when the start angle
+is larger than the end angle.**
+
+Not more than one full rotation is drawn and there is no overlap in drawing for
+angles outside the 0° and 360° range. However, start and end angles can be
+positioned at angles below 0° and above 360°. Through that, and through how wide
+the ColorLine interval is defined, color stops may lie outside the 0° to 360°
```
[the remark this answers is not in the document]
drott: Ok.

OFF_AMD2_WD.md
```diff
@@ -587,61 +587,87 @@ region. Sweep gradients are not bounded, however, and fill the entire space.
...
+of 230°.](images/colr_conic_gradient_start_stop_angles.png)
+
+**Figure 5.30 A sweep gradient with start angle of 110° and an end angle of
+230°, using a `ColorLine` with color stops 0 and 1 for yellow and red and pad
+extend mode.**
+
+NOTE: If the start angle is less than or equal to the end angle, the color line
+progresses from smaller to larger offsets in the counter-clockwise direction
+along the circular arc. If the start angle is larger than the end angle,
+however, the color line is inverted and progresses in the clockwise direction,
```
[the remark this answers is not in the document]
drott: Done.

drott:
@PeterConstable  Note sure why GitHub does not let me comment directly in the thread above, but just wanted to point you to the typo in the edited version:

> NOTE: To allow for a representation of +360°, a bias of 1.0 is used in the representation of ~~stand~~ start and end angles of sweep gradients. [...]

PR #368, opened 2022-10-09
Prepare proposed revisions to WD of OFF 5th Edn related to COLR

PeterConstable:


PeterConstable:
@vlevantovsky: FYI / please review

drott:
Again, thanks, Peter for going through the tedious process of making the OFF amendments and tracking the changes, and preparing the document with proposed changes. I have no objections to merging this as is. But in addition, my hope would be that we have a complete source-form document here or elsewhere based on which we can continue COLR development for future iterations. 

In other words: Can we backport all changes in COLR into this repo's OFF_AMD2_WD.md and/or rename this file to something that makes sense? In any way, the outcome I would like to achieve is that we have a document that is identical to the COLR section in OpenType 1.9.1 in the proposed current form. Alternatively, I am happy to give up this request if OpenType as a whole becomes available in an editable source form to draft and work collaboratively on changes towards it.

Going this way: PR #362 is set-up this way and would update the OFF_AMD2_WD.md file - would it make sense to rename the file and merge that PR, then rebase this on top of such a change?

PS: Are the images here identical to those in #362?

